PR #41, opened 2015-01-02
Added limit and sort functions

alexandrequinto:
Adding functions limit(seriesList, n), sortByTotal(seriesList), sortByMaxima(seriesList), sortByMinima(seriesList), sortByName(seriesList) and its tests.


dgryski:
Merged by hand.

PR #44, opened 2015-02-06
Added function constantLine

avereha:
The step for this function is until - from. The test for it was failing
because of that(step was 0).

I'm not sure I did the right thing by updating the (from, until) pair
from (0,0) to (0,1) in the tests.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -1852,6 +1852,24 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		}
 		return results
 
+	case "constantLine":
+		value, err := getFloatArg(e, 0)
+
+		if err != nil {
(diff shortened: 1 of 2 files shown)


dgryski:
Please revert the test half of this pull request. It's too invasive right now, and not the proper fix.

JaderDias:
This commit cannot be half-reverted. Every single part of it depends on the other.

dgryski:
It was going to be cleaned up by @avereha and resubmitted. The test for constantLine will be moved to its own function,  thus not requiring changing every single other test entry.

JaderDias:
In my opinion the change makes sense, since "until" is exclusive there can be no valid metric request where from = until

PR #51, opened 2015-05-27
Now movingMedian works as graphite does when the window contains undefined values

JaderDias:


dgryski:
LGTM.  Much closer to graphite's output.

PR #58, opened 2015-06-02
Added maximumAbove, maximumBelow, minimumAbove, minimumBelow functions

JaderDias:
functions already present in graphite

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -630,38 +630,53 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		e.target = "averageSeries"
 		return aggregateSeries(e, args, average)
 
-	case "averageAbove", "averageBelow", "currentAbove", "currentBelow": // averageAbove(seriesList, n), averageBelow(seriesList, n), currentAbove(seriesList, n), currentBelow(seriesList, n)
+	case "averageAbove", "averageBelow", "currentAbove", "currentBelow", "maximumAbove", "maximumBelow", "minimumAbove", "minimumBelow": // averageAbove(seriesList, n), averageBelow(seriesList, n), currentAbove(seriesList, n), currentBelow(seriesList, n), maximumAbove(seriesList, n), maximumBelow(seriesList, n), minimumAbove(seriesList, n), minimumBelow
 		args, err := getSeriesArg(e.args[0], from, until, values)
 		if err != nil {
 			return nil
 		}
 
-		n, err := getIntArg(e, 1)
+		n, err := getFloatArg(e, 1)
 		if err != nil {
 			return nil
 		}
 
+		index := strings.IndexAny(e.target, "AB")
+		isAbove := e.target[index:] == "Above"
+		isInclusive := true
+		var compute func([]float64, []bool) float64
+		switch e.target[0:index] {
+		case "average":
+			compute = avgValue
+		case "current":
+			compute = currentValue
+		case "maximum":
+			compute = maxValue
+			isInclusive = false
+		case "minimum":
+			compute = minValue
+			isInclusive = false
+		}
 		var results []*metricData
 		for _, a := range args {
-			switch e.target {
-			case "averageAbove":
-				if avgValue(a.Values, a.IsAbsent) >= float64(n) {
-					results = append(results, a)
+			value := compute(a.Values, a.IsAbsent)
+			if isAbove {
+				if isInclusive {
+					if value >= n {
+						results = append(results, a)
+					}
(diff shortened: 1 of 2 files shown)


expr.go
```diff
@@ -2118,6 +2125,19 @@ func maxValue(f64s []float64, absent []bool) float64 {
 	return m
 }
 
+func minValue(f64s []float64, absent []bool) float64 {
```
dgryski: This is returning the maximum value in the array.  Is the name wrong or is the code wrong?

expr.go
```diff
@@ -641,27 +641,34 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 			return nil
 		}
 
+		re := regexp.MustCompile("^([a-z]+)([AB][a-z]+)$")
```
dgryski: Using a regex here is overkill.  Regexps are fast in Perl because the language is interpreted but the regex engine is in C.  In Go, regexps are slow because the entire language is fast and the regexp engine is complicated.  This string parsing code can be replaced with ` strings.IndexAny` that looks for an uppercase A or B and splits the string at that point.  We know the strings must match because we're in that case already. 

A simple benchmark shows the difference in speed:
```
<dgryski@kaepora[w] ʕ╯◔ϖ◔ʔ╯ > cat main_test.go
package main

import (
	"regexp"
	"strings"
	"testing"
)

func BenchmarkRegexp(b *testing.B) {
	re := regexp.MustCompile("^([a-z]+)([AB][a-z]+)$")
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		re.FindStringSubmatch("currentAbove")
	}
}

func BenchmarkIndex(b *testing.B) {
	for i := 0; i < b.N; i++ {
		strings.IndexAny("currentAbove", "AB")
	}
}
<dgryski@kaepora[w] \ʕ◔ϖ◔ʔ/ > go test -test.bench=.
testing: warning: no tests to run
PASS
BenchmarkRegexp	 1000000	      1195 ns/op
BenchmarkIndex	10000000	       178 ns/op
ok  	github.com/dgryski/w	3.171s
```

You might also find http://commandcenter.blogspot.nl/2011/08/regular-expressions-in-lexing-and.html an interesting read.

expr.go
dgryski: As `maxValue` is just above, it appears the code is wrong.  Perhaps add a test to catch this sort of bug too.

expr.go
```diff
@@ -641,27 +641,34 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
+		}
 		var results []*metricData
 		for _, a := range args {
-			switch e.target {
-			case "averageAbove":
-				if avgValue(a.Values, a.IsAbsent) >= float64(n) {
+			value := compute(a.Values, a.IsAbsent)
+			switch submatch[2] {
+			case "Above":
+				if value >= float64(n) {
```
dgryski: We keep needing to convert n to a float64.  It probably makes sense to have it be a floating point number in the first place.  Change the declaration above for n to `getFloatArg` instead of `getIntArg`.

expr.go
```diff
@@ -641,27 +641,34 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
-					results = append(results, a)
-				}
-			case "currentAbove":
-				if currentValue(a.Values, a.IsAbsent) >= float64(n) {
-					results = append(results, a)
-				}
-			case "currentBelow":
-				if currentValue(a.Values, a.IsAbsent) <= float64(n) {
+			case "Below":
+				if value <= float64(n) {
```
dgryski: Reading the source code for these functions in https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py it appears that the comparision varies between `<=` (or `>=`) and '<' ('>').   That is, we are consistent, but graphite varies. We might want to document this difference in behaviour.

expr.go
JaderDias: Will be done in the next push -f

expr.go
JaderDias: Will be done in the next push -f

expr.go
JaderDias: The test will be modified and the function too, in the next push -f

expr.go
JaderDias: I the last "push -f" I maximumAbove and minimumAbove are exclusive as in the original Graphite. All the 8 methods are now consistent with the python version.

dgryski:
LGTM

PR #59, opened 2015-06-03
Revert "Refactoring common code used by both movingAverage and moving…

JaderDias:
…Median"

This reverts commit 12a8b7178303b091245a9901c4b7a473b5e2f424.

The reverted commit reused code but also hurted the performance of movingAverage significantly
We need to postpone this refactoring until we find a way of doing it without hurting the performance


dgryski:
LGTM, thanks.

PR #60, opened 2015-06-11
movingMedian is now O(log n) instead of O(n) where n is the window size

JaderDias:
also fix a minor error in the median results that was reflected in the automated tests


dgryski:
LGTM

PR #64, opened 2015-07-07
fix a bug on highestCurrent and similar expressions

JaderDias:
highestCurrent and similar expressions didn't handle well a series where all elements are absent


dgryski:
LGTM

PR #71, opened 2015-08-11
Add the isNonNull function.

vision-sbm:
http://graphite.readthedocs.org/en/latest/functions.html#graphite.render.functions.isNonNull

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -935,6 +935,29 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 
 		return results
 
+	case "isNonNull", "isNotNull": // isNonNull(seriesList), isNotNull(seriesList)
+
+		_, err := getSeriesArg(e.args[0], from, until, values)
+
+		e.target = "isNonNull"
+
+		if err != nil {
+			return nil
+		}
+
+		return forEachSeriesDo(e, from, until, values, func(a *metricData, r *metricData) *metricData {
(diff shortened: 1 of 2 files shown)


expr.go
```diff
@@ -935,6 +935,29 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 
 		return results
 
+	case "isNonNull", "isNotNull": // isNonNull(seriesList), isNotNull(seriesList)
+
+		_, err := getSeriesArg(e.args[0], from, until, values)
```
dgryski: Do you even need to call `getSeriesArg()` here?

expr.go
```diff
@@ -935,6 +935,29 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
+
+		e.target = "isNonNull"
+
+		if err != nil {
+			return nil
+		}
+
+		return forEachSeriesDo(e, from, until, values, func(a *metricData, r *metricData) *metricData {
+			for i, _ := range a.Values {
```
dgryski: Don't need the _ here. `for i := range a.Values` will do what you want.

vision-sbm:
Did a rewrite of the lines in question. I was initially worried about not checking for an err, but I see that other existing functions don't take it into consideration either.

dgryski:
LGTM

PR #73, opened 2015-08-13
Fix the format for the timeShift function to include the offset.

vision-sbm:


dgryski:
Do we need to update any unit tests?

vision-sbm:
Interestingly enough, AFAICT there are no tests for timeShift. I'll add some.

dgryski:
Ping.

Issue #77, opened 2015-09-10
Support for Pearson Product Moment Correlation Coefficient (PPMCC)


nnuss:
As mentioned in [FB Gorilla paper](http://www.vldb.org/pvldb/vol8/p1816-teller.pdf) add support for sorting/filtering metrics against a test metric using PPMCC

dgryski:
https://github.com/dgryski/go-onlinestats/blob/master/pearson.go

dgryski:
https://github.com/dgryski/go-onlinestats/blob/master/spearman.go

dgryski:
This is done.

PR #81, opened 2015-09-29
Pearson

nnuss:
Submitted for consideration. Implementation of the PMCC portion of Issue #77
Spearman to follow


nnuss:
Ah, yes that's better. I'll reopen another that will merge cleanly.

PR #82, opened 2015-10-02
Implementation of Pearson product-moment correlation coefficient (PMCC) function(s)

nnuss:
pearson( seriesA, seriesB, windowSize )

   Calculate Pearson score between seriesA and seriesB over windowSize.

pearsonClosest( series, seriesList, n, direction="abs" )

    Return the n series in seriesList with closest Pearson score to the
    first series argument.
    An optional direction parameter may also be given:
        "abs"   - (default) Series with any Pearson score + or - [-1 .. 1].
        "pos"   - Only series with positive correlation score [0 .. 1]
        "neg"   - Only series with negative correlation score [1 .. 0]

Note:
    Pearson (and this implementation) is intolerant of missing values.
    Missing values will result in:
        pearson(): missing values in windows propagate to windows they appear in
        pearsonClosest():
            - series : missing value will result in empty result.
            - seriesList : any series with missing value will be skipped.

    Additionally there is a special case where a series (or window) containing only
    zeros leads to a division-by-zero and will manifest similarly to if
    the series contained a missing value.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -1605,6 +1605,136 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		}
 		return results
 
+	case "pearson": // pearson(series, series, windowSize)
+		arg1, err := getSeriesArg(e.args[0], from, until, values)
+		if err != nil {
+			return nil
+		}
+
+		arg2, err := getSeriesArg(e.args[1], from, until, values)
+		if err != nil {
+			return nil
+		}
+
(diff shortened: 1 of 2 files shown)


nnuss:
Freshly updated PR with e487c3dae420fd08e9c65b6e48a0742d85b87ff8
Allows Pearson to handle the NaN filtering and updates the test (so the NaN-containing metric matches).

PR #86, opened 2015-10-25
Add offset function support

szibis:
http://graphite.readthedocs.org/en/latest/functions.html#graphite.render.functions.offset


dgryski:
LGTM, thanks!

PR #89, opened 2015-10-26
Add perSecond function

szibis:
http://graphite.readthedocs.org/en/latest/functions.html#graphite.render.functions.perSecond


dgryski:
I would prefer this to be its own case in the switch statement.  If you need to refactor the code to eliminate duplicate code, please do so.

szibis:
I make split

dgryski:
Thanks

PR #92, opened 2015-10-29
Add removeBelowValue function.

zimathias:


dgryski:
Thanks.  Any other fixes or features from the lyft branch that should be pushed upstream?

dgryski:
Reverted, sorry.  Totally missed this was the same feature as the lyft PR I commented on re: mutated metrics.

PR #94, opened 2015-10-31
Add changed function

szibis:
http://graphite.readthedocs.org/en/0.9.x/functions.html#graphite.render.functions.changed


dgryski:
Thanks!

PR #98, opened 2015-11-07
Include zero when calculating an integral.

zimathias:


expr.go
```diff
@@ -1215,7 +1215,7 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		return forEachSeriesDo(e, from, until, values, func(a *metricData, r *metricData) *metricData {
 			current := 0.0
 			for i, v := range a.Values {
-				if a.IsAbsent[i] || v == 0 {
+				if a.IsAbsent[i] {
```
zimathias: See the graphite-web integral function:

      if val is None:
        newValues.append(None)
      else:
        current += val
        newValues.append(current)

https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L1114-L1118

dgryski:
LGTM

PR #102, opened 2015-12-02
Pearson closest bug

nnuss:
I found that pearsonClosest() doesn't actually filter closest-n as advertised.
submitted: a test and fix (as well as a tiny change to testing code to get an error rather than a panic)

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -1867,13 +1867,13 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 			heap.Push(&mh, metricHeapElement{idx: index, val: value})
 		}
 
+		if n > len(mh) {
+			n = len(mh)
+		}
 		results := make([]*metricData, n)
-		for len(mh) > 0 {
+		for i := range results {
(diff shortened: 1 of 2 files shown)


dgryski:
Is our top-n stuff broken anywhere else?

nnuss:
By inspection we look ok:
- `{average,current,maximum,minimum}{Above,Below}()` - only have single-value tests but appear correct

- `highest{Average,Current,Max*}()` looks like it could allocate/return some nil series in the face of several metrics with only null - has mutli-value test and otherwise fine &sup1;

- `lowest{Average,Current}()` appears valid for normal operation - has multi-value test. Lacks NaN handling like its highest* cousins which might break things on pathological metrics &sup1;

- `limit()` has multi-value test and is fine

- some other functions such as `remove{Above,Below}Value()` erroneously show up in grepping the code for /, ?n\b/

&sup1; : If I'm being pedantic these functions do not perform the metric ordering when list of series < n. The consequence seems to be that junk series might not get filtered, also that the stacked(...) representation may not be stable/pretty.

Issue #100, opened 2015-11-20
add tukeyBelow(seriesList, basis, b, [interval])


nnuss:
Same logic as tukeyAbove, but for series with outliers below threshold.

dgryski:
Done as https://github.com/dgryski/carbonapi/pull/103

Issue #99, opened 2015-11-12
tukeyAbove(seriesList, interval, basis, n) : interval is unused


nnuss:
[This issue serves as a reminder / todo of interval's inteneded use.]

`interval` was conceived as a mechanism where the quantile threshold can be computed on a (leading IIRC) subset to enable this funciton to more efficiently process large seriesList over large intervals.

@dgryski and I discussed that we could move it to the end as an optional parameter.

dgryski:
Done as https://github.com/dgryski/carbonapi/issues/103

PR #103, opened 2015-12-06
Fixes and features for tukey...

nnuss:
Addresses the following issues (including tests):
* #100 - tukeyBelow()
* #99 - tukeyAbove(): `interval` is unused and should be optional final argument

Also a couple fixes to the use of 'n'[umber] of series to return.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -2351,45 +2351,39 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		}
 		return results
 
-	case "tukeyAbove": // tukeyAbove(seriesList,interval,basis,n)
+	case "tukeyAbove": // tukeyAbove(seriesList,basis,n,interval=0)
 
 		arg, err := getSeriesArg(e.args[0], from, until, values)
 		if err != nil {
 			return nil
 		}
 
-		var n int
-		var scaleByStep bool
-
-		switch e.args[1].etype {
-		case etConst:
-			n, err = getIntArg(e, 1)
-		case etString:
-			var n32 int32
-			n32, err = getIntervalArg(e, 1, 1)
-			n = int(n32)
-			scaleByStep = true
-		default:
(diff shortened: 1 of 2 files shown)


expr.go
```diff
@@ -2351,46 +2351,43 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
-			n = int(n32)
-			scaleByStep = true
-		default:
-			err = ErrBadType
-		}
+		basis, err := getFloatArg(e, 1)
 		if err != nil {
 			return nil
 		}
+		// TODO(nnuss): basis <= 0 ?
```
dgryski: `basis <= 0` should probably be an error (`return nil`)

expr.go
```diff
@@ -2411,7 +2411,10 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		iqr := points[third] - points[first]
 
 		max := points[third] + basis*iqr
-		// min := points[first] - basis*iqr
+		min := points[first] - basis*iqr
+
+		index := strings.IndexAny(e.target, "AB")
+		isAbove := e.target[index:] == "Above"
```
dgryski: `strings.HasSuffix(e.target, "Above")` ?

nnuss:
I was pretty sure about basis but wanted a second opinion; thanks. Without some special handling you kinda-sorta get a "remove pathological series" filter.
Ex. tukeyAbove( metric.*, -1, 1000 ) --> every metric with point >= 1st quartile value.

Also changed isAbove logic to your cleaner HasSuffix() suggestion
(should do that in {current,minimum,maximum}{Above,Below}() also ).

dgryski:
Yeah, when I saw the `strings.IndexAny(e.target, "AB")` line, I thought to myself "I bet I do that too; that's being clever instead of being clear.  I should fix that."

PR #110, opened 2016-02-04
Add support for timeFunction

kerrick-lyft:


dgryski:
Sorry for not reviewing this sooner.  I'm not sure I'm happy with the special casing for the time in the test function.  However, I don't have any better ideas at the moment other than its own custom test function.  We could do something with adding start/stop times to the test struct, but that would involve touching every single test entry.  Dunno...

kerrick-lyft:
OK, want me to move it into its own test function?

dgryski:
Yes,  that's easiest.

kerrick-lyft:
OK, I moved the test to a separate function. There's now some code duplication between the two functions, wasn't sure the best way to deal with it (or if you consider it a problem).

dgryski:
It's not a problem. We can deduplicate the test code later.   Thanks.

Issue #104, opened 2015-12-09
Add exponentialWeightedMovingAverage()


dgryski:
(with shorter name..)

nnuss:
This is a straightforward task for a new contributor.
Here's a test case: nnuss@4761aee769ebbde27abaea4d422ae26e2a26d8a5

PR #112, opened 2016-03-02
add EWMA as per #104

ksurent:


ksurent:
For functions that take more arguments than just seriesList the final target name in the response is "foo(series)" instead of "foo(series,bar)", i.e. the rest of the arguments is not included in the target name.

PR #116, opened 2016-03-09
Compatibility: mostDeviant() arguments changed order in graphite master

nnuss:
See
* graphite-project/graphite-web/pull/279
* graphite-project/graphite-web@44524016474cbdf6d71afb9eb105dea07766d741
... which changed the order of arguments for mostDeviant() on the mainline branch.
The 0.9.x branch retains mostDeviant(n, seriesList).

Grafana expects mostDeviant(seriesList, n).

Accept it either way by detecting the type of the first argument and
changing the logic to "Where is the 'n' arg?" and seriesList is always
"the other one" ( nArg ^ 1 ).

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -1543,13 +1543,20 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 			return min
 		})
 
-	case "mostDeviant": // mostDeviant(n, seriesList)
-		n, err := getIntArg(e, 0)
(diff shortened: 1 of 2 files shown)


nnuss:
Thanks for the comments @dgryski, commented the arg-swapping logic and lightly modified the case comment.

dgryski:
Not sure why I didn't merger this before.

PR #117, opened 2016-03-10
Named arguments

ksurent:
This adds the ability to parse expressions with named arguments, e.g. `summarize(foo.*, '1min', func='max')`. To work with named arguments it adds a bunch of `getTypeNamedOrPosArgDefault()` functions to be used in place of `getTypeArgDefault()` when named arguments support is needed.
Also, I added named arguments to functions that have them according to graphite's documentation.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -422,17 +422,32 @@ func getIntArgDefault(e *expr, n int, d int) (int, error) {
 	return int(e.args[n].val), nil
 }
 
-func getBoolArgDefault(e *expr, n int, b bool) (bool, error) {
+func getBoolNamedOrPosArgDefault(e *expr, k string, n int, b bool) (bool, error) {
 	if len(e.args) <= n {
 		return b, nil
 	}
 
-	if e.args[n].etype != etName {
+	if a := getNamedArg(e.args, k); a != nil {
+		return doGetBoolArgDefault(a, b)
+	}
+
+	return getBoolArgDefault(e, n, b)
+}
+
+func getBoolArgDefault(e *expr, n int, b bool) (bool, error) {
+	if len(e.args) <= n {
+		return b, nil
+	}
+	return doGetBoolArgDefault(e.args[n], b)
+}
+
+func doGetBoolArgDefault(e *expr, b bool) (bool, error) {
+	if e.etype != etName {
 		return false, ErrBadType
 	}
 
 	// names go into 'target'
-	switch e.args[n].target {
+	switch e.target {
 	case "False", "false":
 		return false, nil
 	case "True", "true":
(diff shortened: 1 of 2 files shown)


dgryski:
Eek, I didn't see you had opened this!

I'll review it tonight.

expr.go
```diff
@@ -293,23 +319,40 @@ func getStringArg(e *expr, n int) (string, error) {
...
+		return s, nil
 	}
 
-	return e.args[n].valStr, nil
+	return doGetStringArgDefault(e.args[n], s)
 }
 
-func getStringArgDefault(e *expr, n int, s string) (string, error) {
+func getStringNamedOrPosArgDefault(e *expr, k string, n int, s string) (string, error) {
 	if len(e.args) <= n {
```
nnuss: It's a bit of an edge case but;
In functions with >1 named/optional argument this check may be invalid when one of the "interior" arguments are dropped. Examples of such functions:

- `summarize(seriesList, intervalString, func='sum', alignToFrom=False)`

Functions we have that could be affected:
- `threshold(value, label=None, color=None)` (not supported here yet - but I've implemented privately)
- `timeShift(seriesList, timeShift, resetEnd=True, alignDST=False)` (we don't support alignDST yet)
- `transformNull(seriesList, default=0, referenceSeries=None)` (we don't support referenceSeries yet)

Other graphite functions not supported here yet:
- `linearRegression(seriesList, startSourceAt=None, endSourceAt=None)`
- `sinFunction(name, amplitude=1, step=60)`
- `smartSummarize(seriesList, intervalString, func='sum', alignToFrom=False)`
- `substr(seriesList, start=0, stop=0)`

expr.go
```diff
@@ -334,23 +377,40 @@ func getFloatArg(e *expr, n int) (float64, error) {
...
+}
+
+func doGetFloatArgDefault(e *expr, v float64) (float64, error) {
+	if e.etype != etConst && e.etype != etKV {
 		return 0, ErrBadType
 	}
 
-	return e.args[n].val, nil
+	f := e.val
+	if f == 0 {
```
nnuss: This would make `floatArg=0`, when also not the default such as in `transfromNull()`, impossible to pass. I'm not sure that's a problem or not - but it is new.

expr.go
ksurent: Indeed. The length check is only needed when we know the argument is not a named one. I'll move it from `doGetTypeNamedOrPosArgument` to `getTypeArgDefault`.

ksurent: @nnuss 91455df0 addresses this particular problem

expr.go
ksurent: @nnuss addressed in 8ecaaeb

PR #127, opened 2016-06-28
Make aliasSub compatible with \1 type backreferences.

arodland:
I know this isn't 100% perfect (it doesn't account for the fact that the backslash might be backslashed) but it should work in practical situations, and eliminates the biggest cause of having to rewrite queries.

It rewrites \1 to ${1} and \10 to ${10} in the replacement template before passing it to ReplaceAllString.


dgryski:
thanks!

PR #131, opened 2016-07-08
Add smoothedMovingAverage / smma(seriesList)

nnuss:
An exponentially weighted moving average (see: `ewma(seriesList, alpha)`) where the alpha
component is calculated as 1 / ( n : the number of defined points in
the series ).
## 

I had a private branch of ewma and smma from back in December that I
never merged. This is the smma portion, lightly translated to use the
same library as we did for ewma.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -2237,6 +2237,51 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		}
 		return results, nil
 
+	case "smma", "smoothedMovingAverage": // smma(seriesList)
+		arg, err := getSeriesArg(e.args[0], from, until, values)
+		if err != nil {
+			return nil, err
+		}
+
+		e.target = "smma"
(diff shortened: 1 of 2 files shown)


JaderDias:
I don't think we should have an exponentially weighted moving average where alpha depends on the number of defined points in the series. This would mean different values for values in the beginning of the series whenever we change the "until" time.

dgryski:
Any update?  This needs to be rebased if it's going to be merged.  I agree that the smoothed moving average might be confusing.  Do you have a use case for this?

nnuss:
I implemented it because it was related to ewma. The use case I can locate comes from financial forecasting and uses a smma to smooth the entire range and then a pair of ewma (near-term, medium-term) to signal buy/sell as they cross the smma line.

I favor closing this PR as there has not been interest here, nor in the graphite project, for it.

PR #132, opened 2016-07-08
fix some cases where groupByNode will work incorrectly

Civil:
Test for that case is still TODO. If you don't want to write it by yourself I'll do that on monday.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -1194,13 +1194,19 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 		for k, v := range groups {
 			k := k // k's reference is used later, so it's important to make it unique per loop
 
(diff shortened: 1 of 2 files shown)


Civil:
Example query that'll fail:
```
groupByNode(la.{01,10,15}.value, 1, "sumSeries")
```

Civil:
Fixed commit message, but still haven't added test for that.

Civil:
Found a bug in this fix, will try to get a proper fix...

Civil:
Added two tests that fails.

Civil:
Got an uglier version of fix that's working and passes all the tests.

nnuss:
Looks ok to me. Perhaps later we'll augment parseExpr to hint the etype we want.

PR #136, opened 2016-07-24
Compatibility document moved into the repo

nnuss:
I had been updating a Graphite compatibility document in the wiki section. But keeping that in step with PRs was mostly up to me and not something contributors could readily update. This moves that document to the repository as `COMPATIBILITY.md` - I've also taked the opportunity to update it with recent changes (and also some not so recent ones which I had missed).


dgryski:
LGTM

PR #138, opened 2016-08-01
Add substr() function

arodland:
Another aliasByNode-alike, but different.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -2496,6 +2496,43 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
 			return percentile(values, percent, interpolate)
 		})
 
+	case "substr": // aliasSub(seriesList, start, stop)
+		args, err := getSeriesArg(e.args[0], from, until, values)
+		if err != nil {
+			return nil, err
+		}
+
+		startField, err := getIntNamedOrPosArgDefault(e, "start", 1, 0)
+		if err != nil {
+			return nil, err
+		}
+
+		stopField, err := getIntNamedOrPosArgDefault(e, "stop", 2, 0)
+		if err != nil {
+			return nil, err
+		}
+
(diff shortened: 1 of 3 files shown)


expr.go
```diff
@@ -2496,6 +2496,43 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
+		if err != nil {
+			return nil, err
+		}
+
+		startField, err := getIntNamedOrPosArgDefault(e, "start", 1, 0)
+		if err != nil {
+			return nil, err
+		}
+
+		stopField, err := getIntNamedOrPosArgDefault(e, "stop", 2, 0)
```
nnuss: Comment: by having >1 named argument this function will be subject to the same BUG mentioned in threshold().

expr.go
```diff
@@ -2496,6 +2496,43 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
+		if err != nil {
+			return nil, err
+		}
+
+		fmt.Println(startField, stopField)
```
nnuss: please remove this fmt.Println

expr.go
```diff
@@ -2496,6 +2496,43 @@ func evalExpr(e *expr, from, until int32, values map[metricRequest][]*metricData
...
+
+		var results []*metricData
+
+		for _, a := range args {
+			metric := extractMetric(a.GetName())
+			nodes := strings.Split(metric, ".")
+			if startField != 0 {
```
nnuss: startField and stopField should also check against len(nodes) to avoid "slice bounds out of range".

expr.go
arodland: eek, didn't mean to commit that obviously. :)

nnuss:
Thanks @arodland, lgtm

PR #158, opened 2016-09-14
Add midnight, noon, teatime to dateTimeFromEpoch

liangjacky:


main.go
```diff
@@ -178,11 +185,29 @@ func dateParamToEpoch(s string, d int64) int32 {
 	}
 
 	for _, format := range timeFormats {
-		t, err := time.ParseInLocation(format, s, defaultTimeZone)
+		split := strings.Fields(s)
```
dgryski: What happens if `len(split) > 2` ?

main_test.go
```diff
@@ -0,0 +1,45 @@
...
+	var tests = []struct {
+		input  string
+		output string
+	}{
+		{"midnight", "00:00 1994-Aug-17"},
+		{"noon", "12:00 1994-Aug-16"},
+		{"teatime", "16:00 1994-Aug-16"},
+
+		{"noon 08/16/94", "12:00 1994-Aug-16"},
+		{"midnight 20060816", "00:00 2006-Aug-16"},
```
dgryski: I see the inconsistency here with `midnight 20060816` meaning start of day, but without a date (above) it means end of day.

main.go
```diff
@@ -178,11 +185,29 @@ func dateParamToEpoch(s string, d int64) int32 {
...
+		case "midnight", "noon", "teatime":
+			t, err = time.ParseInLocation(format, ds, defaultTimeZone)
+		default:
+			t, err = time.ParseInLocation(format, s, defaultTimeZone)
+		}
+
 		if err == nil {
+			switch ts {
+			case "noon":
+				t = t.Add(time.Hour * 12)
```
dgryski: As above re: adding hours

main_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+		{"15:04 19940816", "15:04 1994-Aug-16"},
+		{"-1day", "15:30 1994-Aug-15"},
+	}
+
+	for _, tt := range tests {
+		parsedTime := dateParamToEpoch(tt.input, int64(0))
+		ts, err := time.ParseInLocation(shortForm, tt.output, defaultTimeZone)
+		if err == nil {
```
dgryski: If we fail to parse out expected output time from our test, that's a programmer error.  I'm happy to have a `panic()`, here, if you want to be more friendly:
```go
if err != nil {
    t.Fatalf("error parsing time: %q: %v", tt.output, err), 
}
```

This follow the standard "straight-line" code layout for Go: https://medium.com/mat-ryer/line-of-sight-in-code-186dd7cdea88

main.go
```diff
@@ -154,17 +154,24 @@ func dateParamToEpoch(s string, d int64) int32 {
...
-	case "yesterday", "today", "tomorrow":
+	case "yesterday", "today", "tomorrow", "midnight", "noon", "teatime":
 		t := timeNow()
 		switch s {
 		case "yesterday":
-			t.AddDate(0, 0, -1)
-		case "tomorrow":
-			t.AddDate(0, 0, 1)
+			t = t.AddDate(0, 0, -1)
+		case "tomorrow", "midnight":
```
dgryski: Can you check if graphite takes `midnight` to mean start of day (today) or end of day?

main_test.go
```diff
@@ -0,0 +1,45 @@
...
+			actualTime := int32(ts.Unix())
+
+			if parsedTime != actualTime {
+				t.Errorf("Expected %v, got %v", actualTime, parsedTime)
```
dgryski: Since `actualTime` and `parsedTime` are both uint32s, this error message isn't that useful.  
Standard Go test output is:
```go
t.Errorf("dateParamToEpoch(%q, 0)=%v, want %v", tt.intput, parsedTime, actualTime)
```
This is also why it's useful to a name the got/want values as `got` and `want` :)

main_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+		{"noon 08/16/94", "12:00 1994-Aug-16"},
+		{"midnight 20060816", "00:00 2006-Aug-16"},
+
+		{"15:04 19940816", "15:04 1994-Aug-16"},
+		{"-1day", "15:30 1994-Aug-15"},
+	}
+
+	for _, tt := range tests {
+		parsedTime := dateParamToEpoch(tt.input, int64(0))
```
dgryski: s/parsedTime/got/   (see below)
The `0` is a untyped constant, which means you don't need to convert it to an `int64`.  The compiler will do that for you.

https://blog.golang.org/constants

main_test.go
```diff
@@ -0,0 +1,45 @@
...
+		ts, err := time.ParseInLocation(shortForm, tt.output, defaultTimeZone)
+		if err == nil {
+			actualTime := int32(ts.Unix())
```
dgryski: s/actualTime/want/

main.go
nnuss: Trying "midnight" in graphite I get the current date midnight [past] not tomorrow's midnight [future].

main_test.go
```diff
@@ -0,0 +1,45 @@
...
+		return time.Date(1994, time.August, 16, 15, 30, 0, 100, defaultTimeZone)
+	}
+
+	const shortForm = "15:04 2006-Jan-02"
+
+	var tests = []struct {
+		input  string
+		output string
+	}{
+		{"midnight", "00:00 1994-Aug-17"},
```
nnuss: So then my assertion is that "00:00 1994-Aug-16" is expected.

main.go
```diff
@@ -154,17 +154,24 @@ func dateParamToEpoch(s string, d int64) int32 {
...
+			t = t.AddDate(0, 0, 1)
 		}
 
 		yy, mm, dd := t.Date()
 		midnight := time.Date(yy, mm, dd, 0, 0, 0, 0, defaultTimeZone)
+		switch s {
+		case "noon":
+			midnight = midnight.Add(time.Hour * 12)
```
liangjacky: Will it though? I guess what I am trying to say is that 00:00 is the same for all days, DST or not, and noon/teatime are all x hours relative to 00:00.

dgryski: I'm talking specifically about the days when DST changes.  `teatime` is still 4pm, but it's now 15 hours away from midnight.  The other day, when it ticks off there are two 2am hours which means that 4pm is now *17* hours away.

liangjacky:
Should be fixed :)

main.go
```diff
@@ -186,8 +188,11 @@ func dateParamToEpoch(s string, d int64) int32 {
 
 	for _, format := range timeFormats {
 		split := strings.Fields(s)
-		ts, ds := split[0], split[1]
+		if len(split) > 2 {
```
dgryski: Actually, this makes it clear that if `len(split)==1`, we'll panic down below.

Add `{"19940816", "00:00 1994-Aug-16"},` to your list of test cases.

liangjacky: good catch. My code was definitesly breaking for just datestrings.

dgryski:
LGTM

Issue #156, opened 2016-09-13
Support for cactiStyle()


oriordan:
Hi, can you please add support for cactiStyle()?
http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.cactiStyle

Thanks!

dgryski:
@DamianZaremba

nnuss:
Closed with #159

oriordan:
Awesome, thanks guys.

PR #159, opened 2016-09-15
Add cactiStyle function with 'si' support for #156

DamianZaremba:
Adds a cactiStyle function, supporting:

- no unit standardisation or 'si' unit standardisation
- arbitrary unit postfixing

It does not support:

- 'binary' unit standardisation (this should be easy enough to add, but I don't really have test examples for it)

This has been 'roughly' tested, using some internal pre-existing render parameters. Further testing needs to take place, to ensure there are no rounding or other not-so-obvious issues present.

TODO:

- Implement the 'binary' unit standardisation?
- Fix the unit tests

Opening a PR for the purposes of feedback.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -15,6 +15,7 @@ import (
 	"github.com/JaderDias/movingmedian"
 	pb "github.com/dgryski/carbonzipper/carbonzipperpb"
 	"github.com/dgryski/go-onlinestats"
+	"github.com/dustin/go-humanize"
 	"github.com/gogo/protobuf/proto"
 	"github.com/wangjohn/quickselect"
 )
@@ -3266,6 +3267,83 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 		}
 
 		return results, nil
+
+	case "cactiStyle": // cactiStyle(seriesList, system=None, units=None)
+		// Get the series data
+		original, err := getSeriesArg(e.args[0], from, until, values)
+		if err != nil {
+			return nil, err
+		}
+
+		// Get the arguments
+		system, err := getStringNamedOrPosArgDefault(e, "system", 1, "")
+		if err != nil {
+			return nil, err
+		}
+		unit, err := getStringNamedOrPosArgDefault(e, "units", 2, "")
+		if err != nil {
+			return nil, err
+		}
+
+		// Deal with each of the series
+		var metrics []*MetricData
+		for _, a := range original {
+			// Calculate min, max, current
+			//
+			// This saves calling summarizeValues 3 times and looping over
+			// the metrics 3 times
+			//
+			// For min:
+			// Ignoring any absent values and inf (if we have a value)
+			// Using summarizeValues("min", ...) results in incorrect values, when absent
+			// values are present
+			//
+			minVal := math.Inf(1)
+			currentVal := math.Inf(-1)
+			maxVal := math.Inf(-1)
+			for i, av := range a.GetValues() {
+				if !a.IsAbsent[i] {
+					minVal = math.Min(minVal, av)
+					maxVal = math.Max(maxVal, av)
+					currentVal = av
+				}
+			}
+
+			// Format the output correctly
+			min := ""
+			max := ""
+			current := ""
+			if system == "si" {
+				mv, mf := humanize.ComputeSI(minVal)
+				xv, xf := humanize.ComputeSI(maxVal)
+				cv, cf := humanize.ComputeSI(currentVal)
+
+				min = fmt.Sprintf("%.0f%s", mv, mf)
+				max = fmt.Sprintf("%.0f%s", xv, xf)
+				current = fmt.Sprintf("%.0f%s", cv, cf)
+
+			} else if system == "" {
+				min = fmt.Sprintf("%.0f", minVal)
+				max = fmt.Sprintf("%.0f", maxVal)
+				current = fmt.Sprintf("%.0f", currentVal)
+
+			} else {
+				return nil, fmt.Errorf("%s is not supported for system", system)
+			}
+
+			// Append the unit if specified
+			if len(unit) > 0 {
+				min = fmt.Sprintf("%s %s", min, unit)
+				max = fmt.Sprintf("%s %s", max, unit)
+				current = fmt.Sprintf("%s %s", current, unit)
+			}
+
+			r := *a
+			r.Name = proto.String(fmt.Sprintf("%s Current: %s Max: %s Min: %s", *a.Name, current, max, min))
+			metrics = append(metrics, &r)
+		}
+
+		return metrics, nil
 	}
 
 	return nil, fmt.Errorf("unknown function in evalExpr: %q", e.target)
--- a/expr/expr_test.go
+++ b/expr/expr_test.go
@@ -2081,6 +2081,208 @@ func TestEvalExpression(t *testing.T) {
 			[]*MetricData{makeResponse("removeAbovePercentile(metric1, 50)",
 				[]float64{1, 2, -1, 7, math.NaN(), math.NaN(), math.NaN(), math.NaN()}, 1, now32)},
 		},
+		{
+			&expr{
+				target: "cactiStyle",
+				etype:  etFunc,
+				args: []*expr{
+					{target: "metric1"},
+					{valStr: "si", etype: etString},
+				},
+			},
+			map[MetricRequest][]*MetricData{
+				MetricRequest{"metric1", 0, 1}: {
+					makeResponse("metric1",
+						[]float64{math.NaN(), 20531.733333333334, 20196.4, 17925.333333333332, 20950.4, 35168.13333333333, 19965.866666666665, 24556.4, 22266.4, 58039.86666666667}, 1, now32),
+				},
+			},
+			[]*MetricData{
+				makeResponse("metric1 Current: 58k Max: 58k Min: 18k",
+					[]float64{math.NaN(), 20531.733333333334, 20196.4, 17925.333333333332, 20950.4, 35168.13333333333, 19965.866666666665, 24556.4, 22266.4, 58039.86666666667}, 1, now32),
+			},
+		},
+		{
+			&expr{
+				target: "cactiStyle",
+				etype:  etFunc,
+				args: []*expr{
+					{target: "metric1"},
+					{valStr: "si", etype: etString},
+				},
+			},
+			map[MetricRequest][]*MetricData{
+				MetricRequest{"metric1", 0, 1}: {
+					makeResponse("metric1",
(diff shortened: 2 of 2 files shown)


expr/expr.go
```diff
@@ -3266,6 +3267,74 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+		// Get the arguments
+		var system string = ""
+		var unit string = ""
+		if len(e.args) >= 2 {
+			system = e.args[1].valStr
+		}
+		if len(e.args) >= 3 {
+			unit = e.args[2].valStr
+		}
+
```
nnuss: Please make use of the existing `getStringNamedOrPosArgDefault()` function to set these variables. Ex:
```go
        system, err := getStringNamedOrPosArgDefault(e, "system", 1, "")
        if err != nil {
            return nil, err
        }
```

They're mouthfulls, but will allow calls using named params:
* cactiStyle( seriesList, "si", units="firkins")
* cactiStyle( seriesList, units="furlongs")

// With the known limitation that cactiStyle( seriesList, system="si", "fortnights" ) will fail.

expr/expr.go
```diff
@@ -3266,6 +3267,74 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+		// Deal with each of the series
+		for _, a := range original {
+			var min string
+			var current string
+			var max string
+
+			// Calculate min, ignoring any absent values and inf (if we have a value)
+			// Using summarizeValues("min", ...) results in incorrect values, when absent
+			// values are present
```
nnuss: If you adapted the logic present in `case "rangeOfSeries":` you could achieve your goal of the min, max, and current values in a single pass over each series.
Versus 2`*` here:

1. getNotNullMin(a)
1. summarizeValues("max", a.GetValues())

`*` summarizeValues("last",...) doesn't walk the array but also can't check for isAbsent as you noted.

expr/expr_test.go
```diff
@@ -2081,6 +2081,35 @@ func TestEvalExpression(t *testing.T) {
...
+					{target: "metric1"},
+					{valStr: "si", etype: etString},
+					{valStr: "carrots", etype: etString},
+				},
+			},
+			map[MetricRequest][]*MetricData{
+				MetricRequest{"metric1", 0, 1}: {makeResponse("metric1", []float64{1, 2, 3, 4, 5}, 1, now32)},
+			},
+			[]*MetricData{makeResponse("metric1 Current: 5 carrots Max: 5 carrots Min: 1 carrots", []float64{1, 2, 3, 4, 5}, 1, now32)},
+		},
```
nnuss: Suggestions for additional test cases:

1. 1/3
1. 0 +/- e (some very small value)
1. all math.NaN()
1. 1000 exactly
1. slightly more than each of: 10^6, 10^9, 10^12, 10^15
1. similarly: 10^-3, 10^-6, ...

expr/expr.go
DamianZaremba: Done

expr/expr.go
DamianZaremba: I think I've got this into 1 pass now, since the entire array needs to be walked to calculate min, calculating max and last should just work. This assumes the list is in time order for "last", but so does `summarizeValues("last",...)` due to slicing the last element.

expr/expr.go
```diff
@@ -3266,6 +3267,82 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+			val_max := math.Inf(-1)
+			for i, av := range a.GetValues() {
+				if !a.IsAbsent[i] && (val_min == math.Inf(-1) || av < val_min) {
+					val_min = av
+				}
+				if !a.IsAbsent[i] && av > val_max {
+					val_max = av
+				}
+				val_current = av
+			}
```
nnuss: This looks correct. I have some style comments:

* Go projects generally prefer camelCase to snake_case (from golint). val_min -> valMin
* You check `!a.IsAbsent[i]` in both if blocks which trips my **D**on't **R**epeat **Y**ourself sense a little.
* Checking `val_min == math.Inf(-1)` could be more elegantly/readably achieved by initializing `val_min := math.Inf(1)` (+Inf). Then the == check should not be necessary.

There are also math.Min() and math.Max() functions which probably add a small amount of overhead (edge case checking, function call, etc) but aid readability.

With those comments here's what I envision:
```go
    vMin := math.Inf(1)
    vCurrent := math.NaN()
    vMax := math.Inf(-1)
    for i, av := range a.GetValues() {
        if a.IsAbsent[i] {
            continue
        }
        vMin = math.Min(vMin, av)
        vMax = math.Max(vMax, av)
        vCurrent = av
    }
    // error handling if math.IsNaN(vCurrent) meaning we never saw a valid value
```

expr/expr.go
```diff
@@ -3266,6 +3267,82 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+
+			// Append the unit if specified
+			if len(unit) > 0 {
+				min = fmt.Sprintf("%s %s", min, unit)
+				max = fmt.Sprintf("%s %s", max, unit)
+				current = fmt.Sprintf("%s %s", current, unit)
+			}
+
+			a.Name = proto.String(fmt.Sprintf("%s Current: %s Max: %s Min: %s", *a.Name, current, max, min))
```
nnuss: We don't mangle the name in-place (it messes up the tests for one).
Instead follow our convention of:

```go
    //above the for loop
    var results []*MetricData

    for _, a := range original {
        ...
    
        r := *a
        r.Name = proto.String(fmt.Sprintf("%s Current: %s Max: %s Min: %s", *a.Name, current, max, min))
        results = append(results, &r)
    }
    return results, nil
```

expr/expr.go
```diff
@@ -3266,6 +3267,82 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+			val_current := math.Inf(-1)
+			val_max := math.Inf(-1)
+			for i, av := range a.GetValues() {
+				if !a.IsAbsent[i] && (val_min == math.Inf(-1) || av < val_min) {
+					val_min = av
+				}
+				if !a.IsAbsent[i] && av > val_max {
+					val_max = av
+				}
+				val_current = av
```
nnuss: I noticed afterwards that val_current isn't guarded against a.IsAbsent[i] - which caused your test to fail:

```
expr_test.go:2168: bad name for cactiStyle() metric 0: got metric1 Current: 0 Max: 5 Min: 1, want metric1 Current: 5 Max: 5 Min: 1
```

DamianZaremba: Yep - also has an issue with the tests failing due to changed data
`	expr_test.go:2305: cactiStyle(): source data was modified key {metric1 0 1} index 0 want:
		name:"metric1" startTime:1475007879 stopTime:1475007889 stepTime:1 values:0 values:20531.733333333334 values:20196.4 values:17925.333333333332 values:20950.4 values:35168.13333333333 values:19965.866666666665 values:24556.4 values:22266.4 values:58039.86666666667 isAbsent:true isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false 
		 got:
		name:"metric1 Current: 58k Max: 58k Min: 18k" startTime:1475007879 stopTime:1475007889 stepTime:1 values:0 values:20531.733333333334 values:20196.4 values:17925.333333333332 values:20950.4 values:35168.13333333333 values:19965.866666666665 values:24556.4 values:22266.4 values:58039.86666666667 isAbsent:true isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false isAbsent:false 
	expr_test.go:2320: bad name for cactiStyle() metric 0: got metric1 Current: 58k Max: 58k Min: 18k, want metric1 Current:58k Max:58k Min:18k`

Changed to copying the values into a new array and returning that, which seems to satisfy the tests

expr/expr.go
DamianZaremba: See above 👍

expr/expr_test.go
DamianZaremba: Added some of these examples + some random existing known data samples so the output is comparable to the Python impl.

Not sure how many tests this should have, as it seems to be quite verbose in the current setup. Perhaps could re-structure the tests into a loop or test multiple 'metrics' in 1 array entry or maybe just a smaller but more diverse data selection?

expr/expr_test.go
```diff
@@ -2102,13 +2108,159 @@ func TestEvalExpression(t *testing.T) {
...
+					makeResponse("metric1",
+						[]float64{1000}, 1, now32),
+				},
+			},
+			[]*MetricData{
+				makeResponse("metric1 Current: 1k Max: 1k Min: 1k",
+					[]float64{1000}, 1, now32),
+			},
+		},
+		{
```
nnuss: Duplicate test for cactiStyle(metric1,"si") -> "metric1 Current: 1k Max: 1k Min: 1k"

dgryski:
Please squash the commits before merging.

expr/expr.go
```diff
@@ -3341,10 +3342,12 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 				current = fmt.Sprintf("%s %s", current, unit)
 			}
 
-			a.Name = proto.String(fmt.Sprintf("%s Current: %s Max: %s Min: %s", *a.Name, current, max, min))
+			r := *a
+			r.Name =proto.String(fmt.Sprintf("%s Current: %s Max: %s Min: %s", *a.Name, current, max, min))
+			metrics = append(metrics, &r)
```
dgryski: It's more than just the tests.  We only fetch a single metric once, even if it's used in more than one expression.  If we modify a metric value here, then the other expressions will evaluate incorrectly.

nnuss:
lgtm
Thanks @DamianZaremba !

If you want I'd be happy to have an additional test or two that doesn't involve "system":
- cactiStyle(metric1)
- cactiStyle(metric1,units="something")

expr/expr.go
DamianZaremba: Ah, yep... that makes sense :)

DamianZaremba:
Added
`			&expr{
				target: "cactiStyle",
				etype:  etFunc,
				args: []*expr{
					{target: "metric1"},
				},
				namedArgs: map[string]*expr{
					"units": {etype: etString, valStr: "apples"},
				},
			},`
and
`			&expr{
				target: "cactiStyle",
				etype:  etFunc,
				args: []*expr{
					{target: "metric1"},
				},
			},`

As tests without 'system', which pass

DamianZaremba:
Also squashed into 1 commit

expr/expr_test.go
DamianZaremba: Copy paste gets you every time, removed

PR #169, opened 2016-10-10
png: yMax and yMin shouldn't be inf

Civil:
Ignore infinity values while computing yMin and yMax

Otherwise png rendering is broken.


dgryski:
LGTM

PR #177, opened 2016-12-13
Adding new function: polyfit(seriesList, degree=1, offset='0d')

oriordan:
Polynomial regression for basic growth forecasting.


dgryski:
Interesting.  How slow does this end up being for a reasonable query?

I see a couple of obvious optimizations (removing `math.Pow()` calls mostly) but I'm not sure where the bottleneck will actually end up being.  If it's in the matrix code, it probably not worth it yet.

oriordan:
For ~2505600 datapoints:
raw data: ~0m3.065s
polyfit w/ math.Pow: ~0m3.827s
polyfit w/o math.Pow: ~0m3.325s

Issue #66, opened 2015-07-21
Missing holtWinters* functions


JaderDias:
There are people that would like to use with carbonapi the holtWinters functions present in the python version

dgryski:
Use https://github.com/datastream/holtwinters as a base instead of doing the math ourselves,  if possible.

dgryski:
This has stalled slightly because it's not just a case of dropping-in a library.  HoltWinters needs several weeks of earlier data in order to correctly compute the seasonality for its predictions, which means we need to collect bootstrap data similar to how timeShift() works.

dgryski:
I'm replacing the Holt-Winters code and hopefully will fix this at the same time.

dgryski:
The patch in https://github.com/dgryski/carbonapi/commit/d4efd704fa36b876b4318dbba30b32af81068090 can be easily extended to add in the two other HW functions we're missing.

As far as I can tell, nobody is using them at the moment, so I won't implement that right now.

dgryski:
This is sufficiently done.

PR #180, opened 2016-12-16
Add more holtWinters functions

arodland:
This adds `holtWintersConfidenceBands(seriesList, delta=3)` and `holtWintersAberration(seriesList, delta=3)` functions, of use in anomaly detection.

The only one missing is `holtWintersConfidenceArea`. I think that this is implementable — it needs to be rewritten in `Metrics()` to an `areaBetween` that takes `holtWintersConfidenceBands` as `seriesList`. But that's a little subtle for me, and I don't use the cairo renderer anyway, so I leave it up to someone else.

Mostly completes #66.


dgryski:
LGTM.

Thanks!

PR #183, opened 2017-01-10
expr: allow '<' and '>' in metric names

kanatohodets:
This enables support for quoted text searches in carbonsearch,
introduced in
https://github.com/kanatohodets/carbonsearch/commit/6d61f5290b52105137b4f8224c8883d8cbdc909d.

Previously using '<' or '>' was a parse error in render.


dgryski:
LGTM

PR #188, opened 2017-03-10
Adding ProxyHeaders to the list of http hanlers

azhiltsov:
When traffic to carbonapi passing load balancer or any kind of reverse
proxy we cannot see actual client IP in the logs.

This commit trying to address this issue adding handler which inspects
common reverse proxy headers and sets the corresponding fields in the
HTTP request struct. These are X-Forwarded-For and X-Real-IP for the
remote (client) IP address.


dgryski:
LGTM

PR #189, opened 2017-03-10
Add 'prefix' flag to allow manual setting internal metric name prefix

StephenPCG:
This commit added a ``prefix`` flag to allow manual setting internal metric name prefix, so use can better control and organize the internal metrics.


dgryski:
LGTM

Issue #203, opened 2017-04-19
divideSeries differs from graphite-web: requires single dividendSeries when divisorSeries present


cldellow:
The graphite docs suggest that this function can take a series list for the dividends:

```
divideSeries(dividendSeriesList, divisorSeries)
Takes a dividend metric and a divisor metric and draws the division result. 
```

And in fact, in graphite-web, you can do:

```
divideSeries(numerators.*, divisor)
```

and get N series back, where each series is the result of dividing one of the series in `numerators.*` by the `divisor` series.

That's not supported in carbonapi: https://github.com/go-graphite/carbonapi/blob/master/expr/expr.go#L1097 explicitly prohibits it.

Is that intentional? If not, I can try to put together a patch sometime this week. I think it should be possible to support this case while also supporting carbonapi's extended divideSeries use case in https://github.com/go-graphite/carbonapi/commit/e0a9a60d48fe83f7ac685d3bd08fa18b41a88578.

dgryski:
We woul welcome a patch for to enable this functionality.

PR #204, opened 2017-04-20
divideSeries: support seriesList in dividends

cldellow:
Fixes https://github.com/go-graphite/carbonapi/issues/203

For back-compat, introduces a flag to use the old way of naming the new metric
based on the argString; otherwise uses the underlying metric names to generate
the new metric name.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -1073,13 +1073,12 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 			return nil, ErrMissingTimeseries
 		}
 
-		var results []*MetricData
 		firstArg, err := getSeriesArg(e.args[0], from, until, values)
 		if err != nil {
 			return nil, err
 		}
 
-		var useMetricNames = false
+		var useMetricNames bool
 
 		var numerators []*MetricData
 		var denominator *MetricData
@@ -1104,10 +1103,11 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 
 		for _, numerator := range numerators {
 			if numerator.StepTime != denominator.StepTime || len(numerator.Values) != len(denominator.Values) {
-				return nil, errors.New("series must have the same length")
+				return nil, errors.New(fmt.Sprintf("series %s must have the same length as %s", numerator.Name, denominator.Name))
 			}
 		}
 
+		var results []*MetricData
 		for _, numerator := range numerators {
 			r := *numerator
 			if useMetricNames {
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -1073,14 +1073,19 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 			return nil, ErrMissingTimeseries
 		}
 
-		numerators, err := getSeriesArg(e.args[0], from, until, values)
+		var results []*MetricData
+		firstArg, err := getSeriesArg(e.args[0], from, until, values)
 		if err != nil {
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


cldellow:
I'm new to both Go and carbonapi, so please accept my apologies in advance if I'm doing weird things - feedback is definitely welcome.

(and separately, thanks for publishing this tool!)

dgryski:
Thanks! I'll have time to review this tomorrow.

expr/expr.go
```diff
@@ -1073,14 +1073,19 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 			return nil, ErrMissingTimeseries
 		}
 
-		numerators, err := getSeriesArg(e.args[0], from, until, values)
+		var results []*MetricData
```
dgryski: Move this declaration nearer to where it's first used.  (For loop near line 1111 ?)

expr/expr.go
```diff
@@ -1073,14 +1073,19 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+		firstArg, err := getSeriesArg(e.args[0], from, until, values)
 		if err != nil {
 			return nil, err
 		}
 
-		var numerator, denominator *MetricData
-		if len(numerators) == 1 && len(e.args) == 2 {
-			numerator = numerators[0]
+		var useMetricNames = false
```
dgryski: booleans default to false.

`var useMetricNames bool` is sufficient.

expr/expr.go
```diff
@@ -1090,32 +1095,42 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+			denominator = firstArg[1]
 		} else {
 			return nil, errors.New("must be called with 2 series or a wildcard that matches exactly 2 series")
 		}
 
-		if numerator.StepTime != denominator.StepTime || len(numerator.Values) != len(denominator.Values) {
-			return nil, errors.New("series must have the same length")
+		for _, numerator := range numerators {
+			if numerator.StepTime != denominator.StepTime || len(numerator.Values) != len(denominator.Values) {
+				return nil, errors.New("series must have the same length")
```
dgryski: Now that we can have multiple numerators, I wonder if it would be useful to include the name of the 'bad' one in the error message?

expr/expr.go
```diff
@@ -1090,32 +1095,42 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
-		r := *numerator
-		r.Name = fmt.Sprintf("divideSeries(%s)", e.argString)
-		r.Values = make([]float64, len(numerator.Values))
-		r.IsAbsent = make([]bool, len(numerator.Values))
+		for _, numerator := range numerators {
+			r := *numerator
+			if useMetricNames {
+				r.Name = fmt.Sprintf("divideSeries(%s,%s)", numerator.Name, denominator.Name)
+			} else {
+				r.Name = fmt.Sprintf("divideSeries(%s)", e.argString)
```
nnuss: I would prefer the new naming in all cases and have `target=divideSeries(metric.[ab]*)` become named eg. `divideSeries(metric.alice,metric.bob)` rather than remain `divideSeries(metric.[ab]*)` particularly in the case of a bare `*`. Am I in the minority?

cldellow: I personally like the regularity/non-special-caseness of that.

I avoided doing it since it'd be a breaking change to clients who are doing manipulations on the name in graphite expressions, or expecting to read a certain target name in the JSON response -- and I wasn't sure what carbonapi's policy for breaking changes was.

PR #206, opened 2017-04-22
add movingMin/movingMax/movingSum

cldellow:


Civil:
LGTM

expr/expr.go
```diff
@@ -3897,7 +3906,25 @@ func (w *windowed) Stdev() float64 {
 }
 
 func (w *windowed) Mean() float64 { return w.sum / float64(w.Len()) }
+func (w *windowed) Max() float64 {
```
dgryski: Moving min and max are classic interview problems with nice linear-time solutions (see for example http://stackoverflow.com/a/8905575 or https://leetcode.com/problems/sliding-window-maximum/#/solutions ).

It would be interesting to see at what window size using the more complicated data structure makes sense.  For size=5, the loop is probably fine, but if most window sizes are (say) 60, 300, 600 then maybe we should do something smarter here.

I'm fine to merge this, but perhaps make a note that this algorithm could be improved for larger windows.

cldellow: The original graphite-web does a naive approach like I proposed in this PR, but agreed, this could be much faster.

I added a note. Selfishly, my use case is just 5 data points, so I'm OK with the slow approach :)

Issue #205, opened 2017-04-20
applyByNode has limited support


cldellow:
If I understand the code correctly, `applyByNode` has some limitations vs the version available in graphite-web:

1. the `templateFunction` must have only a single argument (due to a bug in https://github.com/go-graphite/carbonapi/blob/68ccaf214d23ca0861a8f20381370fef4f750293/expr/expr.go#L1427)
2. the metrics used in the `templateFunction` must be a subset of the metrics in the initial `seriesList` (...because the list of metrics available to the computation engine are calculated here: https://github.com/go-graphite/carbonapi/blob/68ccaf214d23ca0861a8f20381370fef4f750293/main.go#L412-L440 and Metrics() has no special case for applyByNode: https://github.com/go-graphite/carbonapi/blob/68ccaf214d23ca0861a8f20381370fef4f750293/expr/expr.go#L52
3. the % placeholder can only be used a single time in the `templateFunction`: https://github.com/go-graphite/carbonapi/blob/68ccaf214d23ca0861a8f20381370fef4f750293/expr/expr.go#L1425-L1427
4. the code assumes that the seriesList will be > 1 token long (https://github.com/go-graphite/carbonapi/blob/68ccaf214d23ca0861a8f20381370fef4f750293/expr/expr.go#L1400)

I think this limits much of the power of `applyByNode`, eg, here's a real life query that we use that hits each of these:

```
applyByNode(aliasByNode(servers.com.sortable.ec2.ssp-ue-1.ssp-server.*.dsps.*.imps-offered.count,8),1,"aliasByNode(divideSeries(nonNegativeDerivative(sumSeries(servers.com.sortable.*.*.ssp-server.*.dsps.%.imps-bid-on.count)),nonNegativeDerivative(sumSeries(servers.com.sortable.*.*.ssp-server.*.dsps.%.imps-offered.count))),8)
```

I started trying to make a patch, but ran into issues around the availability of metrics once the expression was rewritten.

I ultimately concluded that `applyByNode` doesn't fit into the current `EvalExpr` model.

Perhaps there should be a query expansion/rewriting phase in `renderHandler` before invoking `EvalExpr`? Rationale: `applyByNode` is sort of a meta-function that exists to produce new targets. In particular, it doesn't need to know the values of any of its series to do its work, just needs to know which metrics exist on the server.

[...]

PR #207, opened 2017-04-23
Allow applyByNode to generate new targets

cldellow:
https://github.com/go-graphite/carbonapi/issues/205

Opening PR for discussion + feedback. This fixes some limitations in the current applyByNode code, but feels a bit hacky.

In particular:

- `RewriteExpr` ends up calling `EvalExpr` (via `getSeriesArg`) outside of
  the defer block
  - I'm not sure what the ramifications of this are.
- support for the `newName` parameter is achieved via stuffing an `alias` arg in
  - I guess I could return the new target name and teach the render function
    to use that instead?

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -1387,7 +1387,7 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 
 		return args, nil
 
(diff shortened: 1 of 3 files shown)


dgryski:
Sorry I haven't looked at this yet.  I'm on vacation at the moment and watching my kids is taking all my spare time.

dgryski:
Sorry for the delay.  This looks fine for a start.  Lets merge it and see how it plays out in production.

PR #214, opened 2017-05-11
Make asPercent compatible with Graphite

Civil:
Related #213 

Please review that PR and if it's OK, I'll do the same for divideSeries.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -738,7 +738,7 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 		var getTotal func(i int) float64
 		var formatName func(a, b string) string
 		var totalString string
-		multipleSeries := false
+		var multipleSeries bool
 
 		if len(e.args) == 1 {
 			getTotal = func(i int) float64 {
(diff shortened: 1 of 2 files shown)


Civil:
@nnuss @JaderDias @dgryski

dgryski:
I'll review this today.

expr/expr.go
```diff
@@ -736,7 +736,9 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 		}
 
 		var getTotal func(i int) float64
-		var formatName func(a *MetricData) string
+		var formatName func(a, b string) string
+		var totalString string
+		multipleSeries := false
```
dgryski: `var multipleSeries bool`

expr/expr.go
```diff
@@ -781,44 +784,94 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
-		for _, a := range arg {
-			r := *a
-			r.Name = formatName(a)
-			r.Values = make([]float64, len(a.Values))
-			r.IsAbsent = make([]bool, len(a.Values))
-			results = append(results, &r)
-		}
+		if multipleSeries {
+			for i := range e.args {
+				if i >= len(e.args)/2 {
```
dgryski: `len(e.args)/2` -- this seems confusing.

Can you document why this is?  It might be nicer to reslice args so you have
```
numerators := e.args[:len(e.args)/2]
denominators := e.args[len(e.args)/2:]
```
and then range over the `numerators` slice.

expr/expr.go
```diff
@@ -781,44 +784,94 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
-				r := results[j]
-				a := arg[j]
+				total := getTotal(i)
 
-				if a.IsAbsent[i] || math.IsNaN(total) || total == 0 {
-					r.Values[i] = 0
-					r.IsAbsent[i] = true
-					continue
-				}
+				for j := range results {
```
dgryski: `for j, r := range results {`

Civil: I've decided to leave that as is for 2 reasons:
1. If I modify stuff like that I still need to create a new "r" variable. Otherwise I got an error "source data was modified" during tests.
2. It's done in the same way as for "b" which makes it easier to read.

dgryski:
LGTM

dgryski:
Oops, sorry forgot to merge this.  I though @Civil would push it.

Civil:
I wanted to do the same for divideSeries, but got distracted by SRECon. Also I found another case that's not handled by this PR - it doesn't handle asPercent(foo*, bar*), but allows you to do asPercent(foo1, foo2, bar1, bar2) which I think is not the graphite-web's behavior.

PR #217, opened 2017-05-16
Use request context instead of context.Background()

amenzhinsky:


deniszh:
Travis says that will not gonna work. Also some commentary why do you want to do this will be nice too.

amenzhinsky:
@deniszh I'm sorry, I forgot to delete unused import.

So this PR makes stopping http server more consistent because when you stop it all request context are cancelled as well. 

At the same time we're working on the autorization feature for carbonapi and it would be nice to use request contexts since we use handler wrappers like `authUser(renderHandler)`.

Thanks.

amenzhinsky:
CI build is failing, do you have any idea why?

Civil:
Seems to be a bug in build script, it shouldn't try to build packages for PR requests.

deniszh:
Upload to packagecloud is broken. Not sure that we should do that on every Travis run, though.
@Civil ?

PR #222, opened 2017-05-29
add a new expression: removeBetweenEpochs

JaderDias:
When summarizing over a long time period

    summarize(metric1, '1d')

sometimes we need to remove an interval where our data is wrong.

    summarize(removeBetweenEpochs(metric1, 1496079025, 1496079115), '1d')

so if you know that your graphite metrics are broken for a specific
interval, you don't need to rewrite their history or delete it entirely.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -3613,6 +3613,48 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 
 		return results, nil
 
+	case "removeBetweenEpochs": // removeBetweenEpochs(seriesLists, fromEpoch, untilEpoch)
+		args, err := getSeriesArg(e.args[0], from, until, values)
+		if err != nil {
+			return nil, err
+		}
+
+		fromEpoch, err := getIntArg(e, 1)
+		if err != nil {
+			return nil, err
+		}
+
+		untilEpoch, err := getIntArg(e, 2)
+		if err != nil {
+			return nil, err
+		}
+
+		var results []*MetricData
+
+		for _, a := range args {
+			r := *a
+			r.Name = fmt.Sprintf("%s(%s, %d, %d)", e.target, a.Name, fromEpoch, untilEpoch)
+			r.IsAbsent = make([]bool, len(a.Values))
+			r.Values = make([]float64, len(a.Values))
+
+			epoch := int(a.StartTime)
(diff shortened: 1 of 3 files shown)


expr/expr.go
```diff
@@ -3613,6 +3613,48 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 
 		return results, nil
 
+	case "removeBetweenEpochs": // removeBetweenEpochs(seriesLists, fromEpoch, untilEpoch)
+		args, err := getSeriesArg(e.args[0], from, until, values)
+		if err != nil {
+			return nil, err
+		}
+
+		fromEpoch, err := getIntArg(e, 1)
```
dgryski: Maybe support epochs in whatever format we accept input dates in?  That would require moving `dateParamToEpoch` from the server package to the expr package.  That would also allow the epochs to be `int32` which would remove a bunch of conversion to `int`s.

JaderDias: Very good suggestion, but the fact that dateParamToEpoch uses Config.defaultTimeZone forces us to make expr.go stateful. That implies that we have to convert some of its static functions into object methods.

nnuss:
I have a more general comment that this approach only works for a single request. If the metric has bad data I think we would want the ability for all requests to know a given interval was bad.

Something that would fit in the current ecosystem would be having a masking series.
- Original data is unmodified
- The masking series would be null except for periods where the original series contains data we wish to mask. In those cases the value is 1.
- Perhaps `maskSeries( original.series, mask.original.series )`
- This makes it relatively easy to mask several time periods
- ! Care would be needed such that the masking series and original series have the same resolution and retention bands (agg method `max` and x-files factor `0.5` ?).

JaderDias:
@nnuss it's technically possible to achieve it without creating a new expression. removeAboveValue(max(mask.original.series, original.series), maxFloat-1) where mask.original.series is either null or maxFloat for any given point.

JaderDias:
@nnuss @dgryski @Civil  I created https://github.com/go-graphite/carbonapi/issues/223 to discuss this feature

JaderDias:
This pull request has been superseded by https://github.com/go-graphite/carbonapi/pull/224

dgryski:
Closing as per https://github.com/go-graphite/carbonapi/pull/222#issuecomment-304888950

PR #229, opened 2017-07-21
add cumulative

errx:
http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.cumulative

https://github.com/graphite-project/graphite-web/blob/36ec37729ca9c4de58127f2177c33a572806666b/webapp/graphite/render/functions.py#L1389


Civil:
LGTM

PR #230, opened 2017-07-26
Linear regression

errx:


errx:
fix linearRegression test

PR #232, opened 2017-08-01
add groupByNodes

errx:
http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.groupByNodes


nnuss:
LGTM also. Thanks @errx 
Graphite handles it by dying also - but can we give a better error when a node number is out of bounds ?

Issue #231, opened 2017-08-01
[Feature Request] Implement divideSeriesLists() function


timtofan:
Please implement divideSeriesLists() function.

nnuss:
Reference docs:
http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.divideSeriesLists

Graphite implementation:
https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L619

towolf:
Yeah, needed this today too.

Civil:
Seems to fixed by PR #233

PR #233, opened 2017-08-15
add divideSeriesLists

errx:
http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.divideSeriesLists

#231


Civil:
It seems to me that test case doesn't cover multiple metrics in request.

Civil:
ah sorry, my bad. Seems to be ok.

PR #235, opened 2017-08-16
carbonapi shouln't fail if some series in expression args doesn't exist

### LINKED ISSUE #90: carbonapi handles missing metrics differently to graphite
If you specify a missing metric as part of a larger formula, the query will fail in carbonapi but succeed in Graphite.

A simple example of this is:

```
data.metric1  # Exists
data.metric2  # Does not exists

diffSeries(data.metric1, data.metric2) # Fails on carbonapi
diffSeries(data.metric1, transformNull(data.metric2, 0)) # Also fails on carbonapi
```

I would expect the result to be `data.metric1` since I'd assume that `data.metric1 - [undefined]` would equal `data.metric1`.

A valid use case in this situation is as followed:

```
# Metric layout
page.[page_name].requests
page.[page_name].errors

# The number of total successful pages served (number of requests - number of errors)
# If we automate this in any form we may hit a page that has never had an error recorded.
diffSeries(page.index_html.requests, page.index_html.errors) 
```

Civil:
This should make carbonapi's behavior compatible with graphite-web/graphite-api

Fixes #90

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -533,13 +533,22 @@ func getSeriesArg(arg *expr, from, until int32, values map[MetricRequest][]*Metr
 	return a, nil
 }
 
+func removeEmptySeriesFromName(args []*MetricData) string {
+	var argNames []string
+	for _, arg := range args {
+		argNames = append(argNames, arg.Name)
+	}
+
+	return strings.Join(argNames, ",")
+}
+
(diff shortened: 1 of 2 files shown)


Civil:
I don't like the idea of having "if len(args) < len(e.args) {" all over the place, but I'm not sure that there is more elegant way to rewrite argString for diffSeries.

deniszh:
Well. I don't know either TBH.

PR #236, opened 2017-08-19
Update to keep up with go dep changes

gothrek22:
Hi,

this is a really short PR to accomodate changes done in the PR below in go dep.

https://github.com/golang/dep/pull/644

I've tested it on my setup and it seems to work properly.


Civil:
Thanks!

PR #239, opened 2017-09-06
Add ability to be configured by ENV variables

ctrlok:
as part of https://github.com/go-graphite/carbonzipper/issues/51


Civil:
You also need to add 'viper' to Gopkg.toml and do vendor tree.

I'll have a more detailed look tomorrow

PR #247, opened 2017-10-13
graphutil: Swap red and darkred

ibuclaw:
This is as per graphite-api.


Civil:
Hey, I've actually decided to make things a little bit different in last release of carbonapi (the one that's now "master"). By default it will have swapped colors, but in config you can actually redefined default values.

I hope this will keep graphs nice by default but will allow to have correct behavior as in graphite-api if you need that.

ibuclaw:
Sure, no problem.

PR #248, opened 2017-10-13
Correctly set name field for toplevel completer queries

ibuclaw:
So that both `name` and `path` fields are set when you request `/metrics/find?format=completer&query=*`.

This matches the behaviour of graphite-web.


Civil:
LGTM

ibuclaw:
Thanks!

PR #251, opened 2017-10-17
Adding fallbackSeries from graphite-api

gksinghjsr:
Adding support for fallbackSeries from graphite api.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -1365,6 +1365,23 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 		}
 		return results, nil
 
+	case "fallbackSeries": // fallbackSeries( seriesList, fallback )
+		/*
+			Takes a wildcard seriesList, and a second fallback metric.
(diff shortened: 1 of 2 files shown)


expr/expr.go
```diff
@@ -1365,6 +1365,23 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
...
+			Takes a wildcard seriesList, and a second fallback metric.
+			If the wildcard does not match any series, draws the fallback metric.
+		*/
+		seriesList, err := getSeriesArg(e.args[0], from, until, values)
+		fallback, errFallback := getSeriesArg(e.args[1], from, until, values)
+		if errFallback != nil && err != nil {
+			return nil, errFallback
+		}
+
+		if seriesList != nil && len(seriesList) > 0 {
```
deniszh: Maybe worth to change to
```if seriesList != nil && len(seriesList) > 0 {
    return seriesList, nil
}
return  fallback, nil```
?

gksinghjsr: Indeed!

PR #252, opened 2017-10-17
Adding support for fallbackSeries from graphite-api

gksinghjsr:
Implemented the fallbackSeries from classical graphite implementation.


Civil:
LGTM. The only thing I'd change is to add test following 2 cases:
1. metric1 doesn't exist
2. metric2 doesn't exist

PR #256, opened 2017-10-30
cairo: Add lineWidth function and support for rigthWidth/leftWidth

ibuclaw:
Implemented `lineWidth()`, and noticed that `rightWidth` and `leftWidth` were being ignored also.  Fixed both cases.

### DIFF
--- a/expr/cairo.go
+++ b/expr/cairo.go
@@ -35,6 +35,7 @@ type graphOptions struct {
 	secondYAxis    bool
 	dashed         float64
 	hasAlpha       bool
+	hasLineWidth   bool
 	stacked        bool
 	stackName      string
 }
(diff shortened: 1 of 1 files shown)


expr/cairo.go
```diff
@@ -2258,7 +2283,9 @@ func drawLines(cr *cairoSurfaceContext, params *Params, results []*MetricData) {
 			clipRestored = true
 		}
 
-		cr.context.SetLineWidth(params.lineWidth)
+		if series.hasLineWidth {
+			cr.context.SetLineWidth(series.lineWidth)
```
ibuclaw: Note, this is reset back to `originalWidth` at the bottom of this loop scope.

PR #259, opened 2017-10-31
Parse "m" as minutes

borovskyav:
Hi! Graphite-web can parse "m" symbol, for example in summarize function as minute unit. I want to add this feature in carbon-api too for backward compatibility.

### DIFF
--- a/expr/helper.go
+++ b/expr/helper.go
@@ -41,7 +41,7 @@ func IntervalString(s string, defaultSign int) (int32, error) {
 		switch unitStr {
 		case "s", "sec", "secs", "second", "seconds":
 			units = 1
-		case "min", "mins", "minute", "minutes":
+		case "m", "min", "mins", "minute", "minutes":
 			units = 60
 		case "h", "hour", "hours":
 			units = 60 * 60


ibuclaw:
> Hi! Graphite-web can parse "m" symbol, 

I don't think that has ever been the case.

https://github.com/graphite-project/graphite-web/blob/c5dc26bb7d88989dd6c6dcd6ac8e875ee73d9a46/webapp/graphite/render/attime.py#L182-L191

And it looks like this breaks one of the TestInterval tests.

errx:
this commit breaks tests

--- FAIL: TestInterval (0.00s)
	helper_test.go:46: intervalString("10m10s")=610, want 0
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x13baa3d]

deniszh:
Yep, I think it should be reverted.

deniszh:
Ah, @errx fixed the tests, OK then.

ibuclaw:
Should probably still be reverted though.  `summarize(foo, '10m')` does not work in graphite-web.

azhiltsov:
I am also in favor of revert, as we are targeting to be original graphite stack compatible
What is your thought @Civil

ibuclaw:
Ping @Civil

Civil:
Compatibility means that every graphite-web's request should be valid inside carbonapi, but we are not aiming to be 100% backward compatible with graphite-web.

So the PR will stay merged.

Civil:
@borovskyav btw, to solve the argue about backward compatibility with graphite - can you please spend some time and do the PR to graphite-web to implement the same functionality? :)

AlexAkulov:
@deniszh maybe we should add 'm' to graphite-web back?

deniszh:
I'm OK with that, please make PR - we'll merge it if no objections.

PR #265, opened 2017-11-23
Add unicode support

korservick:
Some fixes to support unicode simbols in metrics name


expr/expr.go
```diff
@@ -4128,29 +4128,49 @@ func alignToBucketSize(start, stop, bucketSize int32) (int32, int32) {
 	return start, newStop
 }
 
-func extractMetric(m string) string {
+func extractMetric(s string) string {
 
 	// search for a metric name in `m'
```
Civil: please change the variable name in the comments.

korservick:
Fixed

PR #267, opened 2017-11-24
Structure access logging and handler tests

gksinghjsr:
1.Structured access logging to ensure that access logs are standardized. Uses protobuf3.
2. Add tests for Info, Render and Find handlers. The carbon zipper api is mocked.
3. Also incorporates changes suggested in the last pull request.


Civil:
LGTM

PR #268, opened 2017-11-24
fix tests

errx:


errx:
fixing https://github.com/go-graphite/carbonapi/pull/259

PR #270, opened 2017-11-27
multiplySeriesWithWildcards and stddevSeries

kamaev:
added graphite functions


expr/expr.go
```diff
@@ -1382,7 +1382,9 @@ func EvalExpr(e *expr, from, until int32, values map[MetricRequest][]*MetricData
 					if hasVal[i] == false {
 						r.Values[i] = v
 						hasVal[i] = true
-					} else {r.Values[i] *= v}
+					} else {
+						r.Values[i] *= v
+						}
```
Civil: Please run `go fmt` on that file.

kamaev: Sorry! Haven't noticed. I've just fixed it

PR #272, opened 2017-12-17
Separate MarshalPNG/CSV that takes params

borovskyav:
Hi! In [Moira](https://github.com/moira-alert) we need to render trigger timeSeries graphs for showing stats to user, for example, in notification. But it means that we can't use MarshalPNG with http.Request argument. In this PR i have made public fields in params struct, rename methods MarshalCSV(PNG) -> MarshalCSV(PNG)Request and create two methods that have Params argument.

### DIFF
--- a/expr/cairo.go
+++ b/expr/cairo.go
@@ -553,50 +553,50 @@ type Area struct {
 }
 
 type Params struct {
-	width      float64
-	height     float64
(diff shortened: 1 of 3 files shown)


borovskyav:
Have another idea to resolve my problem. Will realize it and create new PR

PR #279, opened 2018-01-19
Added compatibility for mapSeries

kipwoker:
Hello.

According to compatibility page I've made implementation of mapSeries. Check it out please. Also I have commit for reduceSeries. Am I right it should be separate PR?


Civil:
Yes, it's better to have separate pr for reduce

PR #283, opened 2018-01-31
Fix integer overflow when building for 32 bit

alexeypetrenko:
When building for `arm` it fails with error:
```
expr/picture_params.go:394:16: constant 4294967295 overflows int
```

Probably it should be `math.MaxInt32` instead of `math.MaxUint32`


dgryski:
LGTM

PR #284, opened 2018-02-01
Upgraded asPercent to latest version (1.1.1)

kipwoker:
Added logic for 'nodes' parameter.


kipwoker:
Yeah. Actually just added piece with this condition https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L828

PR #286, opened 2018-02-02
Fix frange rounding errors when generating ylabels.

ibuclaw:
For example, when `yMin=0`, `yMax=7` and `yStep=1.2`, the calculated `yTop=7.199999999999999`

How it is worked out is `1.2 * Ceil(7/1.2)`.

This results in frange returning (0.0 .. 6.0) instead of the expected (0.0 .. 7.2).

### DIFF
--- a/expr/cairo.go
+++ b/expr/cairo.go
@@ -363,6 +363,13 @@ var xAxisConfigs = []xAxisStruct{
 	},
 }
 
+// We accept values fractionally outside of nominal limits, so that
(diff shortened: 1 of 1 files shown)


ibuclaw:
> This const should be called floatEpsilon and be unexported.

I looked up on how one would unexport a symbol in Go.  On finding out, I must admit it did raise a few eyebrows.  But oh well, not my language to worry about. :-)

PR #289, opened 2018-02-07
Refactor expression parser

Civil:
This is first part of the commit.

1. Splits expression parser (the one that splits functions, etc) to a separate package, so it's now reusable by other code.
2. Add bunch of helper functions that allows to create new expressions in simpler way


Civil:
Ok, packaging fails but it seems to be unrelated to refactoring

PR #300, opened 2018-03-16
use mapstructure for viper

errx:
viper uses mapstructrure prefix not yaml for name tags


errx:
Also it will fix `tz` setting

Civil:
For some reason tests are not passing anymore after this PR. Can you please fix that?

Civil:
Not sure why, but travis was trying to verify previous version of commit for some reason.

PR #304, opened 2018-03-19
add log directory and logrotate config

ihard:
Add config for logrotate.


ihard:
I do not really understand how my changes could affect the tests in which the application does not see its config ...

Civil:
There is something strange with the tests in master, I'm not sure why they are not passing, while they are ok on a clean environment outside of travis.

Issue #305, opened 2018-03-20
panic using reduce\mapSeries functions


nikita-b:
Hi! I have the problem with last version carbonapi 0.10. When I request this metric:
```
reduce(mapSeries(stats.gauges.production.dedicated.*.master.rooms.*, 4), 'diffSeries', 7, 'limit', 'busy')
```
I get this error:
```
[...]

PR #306, opened 2018-03-20
Skip metrics with missing pair in reduceSeries

errx:
related to https://github.com/graphite-project/graphite-web/issues/1516

Currently if reduceSeries can't find pair for matcher it will panic. Probably a better way is to skip completely metrics without pair.


Civil:
I think it also fixes #305

PR #317, opened 2018-05-18
Gmagnusson/time shift

gunnihinn:
We noticed that `timeShift` had accidentally been changed to be the constant function that returns an empty set of metrics. These patches fix that.

### DIFF
--- a/expr/functions/sum/function_test.go
+++ b/expr/functions/sum/function_test.go
@@ -1,0 +1,49 @@
+package sum
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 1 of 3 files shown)


dgryski:
When was this bug introduced?

gunnihinn:
@dgryski As noted in the commit message for [4fd1124](https://github.com/go-graphite/carbonapi/pull/317/commits/4fd1125b29ceaff50e65a5a30525dd176c3ec88e), it was introduced in [70146ac](https://github.com/go-graphite/carbonapi/commit/70146ace10e1aa702cf0f238d8fcfd4c12bca972).

PR #344, opened 2018-07-19
Booking

gunnihinn:
This pull request contains a large dump of commits we've been making at Booking in an attempt to make carbonapi happy with its life. The commits in this request fall largely into two topics:

**Commits 720cd9e0 to 56ed5c5e**
We had a lot of problems whose root was the complicated recursive call tree that making any request would generate. The new config format is parsed into a tree structure of ServerClient interfaces, and we would recursively fan out requests to the leaf nodes of that tree. This had numerous problems, which for example lead to us making exponential Find request amplification attacks on ourselves and race conditions when merging responses up the call tree. In these commits we get around most of these problems by linearizing the tree into a simple list of leaf nodes before we make any call, and simply call each leaf concurrently.

**Commits 265f83c0 to df9bad43**
These are essentially a linter and static analyzer rampage.


zipper/broadcast/broadcast_group.go
```diff
@@ -102,7 +102,7 @@ func fetchRequestToKey(prefix string, request *protov3.MultiFetchRequest) string
 	return string(key)
 }
 
-func (bg *BroadcastGroup) doSingleFetch(ctx context.Context, logger *zap.Logger, client types.ServerClient, request *protov3.MultiFetchRequest, doneCh chan<- string, resCh chan<- *types.ServerFetchResponse) {
+func (bg *BroadcastGroup) doSingleFetch(ctx context.Context, logger *zap.Logger, client types.ServerClient, requests []*protov3.MultiFetchRequest, doneCh chan<- string, resCh chan<- *types.ServerFetchResponse) {
```
Civil: I do not really understand why it was required to move for-loop here. Also I think that in that case 'doSingleFetch' must be renamed to something else, as it's not a single MultiFetchRequest anymore. But that can be done after PR is merged.

zipper/types/response.go
```diff
@@ -135,12 +135,34 @@ func mergeFetchResponsesWithEqualStepTimes(m1, m2 *protov3.FetchResponse, uuid s
 	return nil
 }
 
+func mergeFetchResponsesWithUnequalStepTimes(m1, m2 *protov3.FetchResponse, uuid string) error {
```
Civil: Again, this can be fixed after this PR, but my opinion on merge logic is that we should prefer metric with less NaNs - because it might be possible that on a one server metric was created with 1s:7d but on other - with 60s:30d - and if we see a lot of NaN in the most precise one, but not in other - we should just drop the most precise one.

I'd also suggest to look into how it's done in Metrictank for consolidation: https://github.com/grafana/metrictank/blob/a673544664502e47ec25798b323c56f3daaccf3a/consolidation/consolidate.go

zipper/types/response.go
```diff
@@ -117,74 +117,50 @@ func swapFetchResponses(m1, m2 *protov3.FetchResponse) {
 	m1.StopTime, m2.StopTime = m2.StopTime, m1.StopTime
 }
 
-func MergeFetchResponses(m1, m2 *protov3.FetchResponse, uuid string) *errors.Errors {
```
Civil: The general cases that happens quite often are:
1. There is a different size because clock is different and we have one extra point in one of the metrics. It actually must be addressed as it's one of the most annoying one.
2. Second case where start point differs, it's quite rare though, but might happen with different storages. Also should be addressed.
3. Third case is where steps differs, in that case logic should be more clever, but see comment in next commit about that. It should be mixture of the one that's in Graphite now and some clever heuristics - we should mark as deffective metrics where there are a lot of NaNs. Maybe have a logic to determine real timestep for data and prefer it.

zipper/zipper_test.go
```diff
@@ -154,35 +154,6 @@ func TestMergeValues(t *testing.T) {
 
 			expectedError: errors.Errors{},
 		},
-		{
```
Civil: I disagree with the reasons behind this commit. This is an OpenSource project and one of the burdens that comes with it - you must take into account what community wants. So I'll consider this as a regression that blocks this PR.

Merge logic *must* take into account different situations for the merging, as it's one of the most annoying problems outside B

Makefile
```diff
@@ -10,7 +10,7 @@ GO ?= go
 PKG_CARBONAPI=github.com/go-graphite/carbonapi/cmd/carbonapi
 PKG_CARBONZIPPER=github.com/go-graphite/carbonzipper/cmd/carbonzipper
 
-carbonapi:
+carbonapi: $(shell find . -name '*.go' | grep -v 'vendor')
```
Civil: Not sure what's this change is about

zipper/types/response.go
```diff
@@ -35,7 +35,7 @@ func (first *ServerInfoResponse) Merge(second *ServerInfoResponse) *errors.Error
 	}
 
 	if first.Err == nil {
-		first.Err = &errors.Errors{}
+		first.Err = new(errors.Errors)
```
Civil: I'd say use of "new" is generally discouraged and this should be either consistent (e.x. https://dave.cheney.net/2014/08/17/go-has-both-make-and-new-functions-what-gives ) or documented and have very special meaning.

cmd/carbonapi/main_test.go
```diff
@@ -9,25 +9,15 @@ import (
...
-	logger      *zap.Logger
-	statsSender func(*zipperTypes.Stats)
-}
+type mockCarbonZipper struct{}
 
 func newMockCarbonZipper() *mockCarbonZipper {
-	z := &mockCarbonZipper{}
-
-	return z
+	return new(mockCarbonZipper)
```
Civil: Same about new as in previous comment.

Makefile
```diff
@@ -19,7 +19,7 @@ debug:
 nocairo:
 	$(GO) build -ldflags '-X main.BuildVersion=$(VERSION)' $(PKG_CARBONAPI)
 
-carbonzipper:
+carbonzipper: $(shell find . -name '*.go' | grep -v 'vendor')
```
Civil: Same question as in one of the previous commits - I don't understand the reason behind this change.

zipper/types/response.go
gunnihinn: If `type A struct{ //... }` then `new(A)` has the same meaning as `&A{}`. It returns a pointer to initialized memory. I agree that if the struct has fields we also want to initialize at the same time, then writing
```
a := &A{
    foo: 10,
    bar: 20,
}
```
is better than:
```
a := new(A)
a.foo = 10
a.bar = 20
```
Here, the zero values for the fields in `Errors` are the only ones we'd want.

> I'd say use of "new" is generally discouraged

I haven't seen that. If you know of discussions where that conclusion is reached I'd be interested in seeing them. Cheney is explaining the difference between two language keywords. `new` gives us a non-nil pointer to a type and it seems fine to use it for that.

Makefile
gunnihinn: It declares all non-vendored `*.go` files in the project to be prerequisites for the `carbonapi` target (see [the make manual]) This means that if any Go file in the project changes, then writing `make carbonapi` will rebuild the binary. If we omit the prerequisites, `make carbonapi` will not rebuild the binary if it already exists, even if the existing binary was built from source files that have now changed. For the `carbonapi` and `carbonzipper` targets, this is handy for checking they still compile when making new changes.

[the make manual]: https://www.gnu.org/software/make/manual/make.html#Rule-Example

zipper/types/response.go
Civil: I'll try to find you a link, but even in the article I gave you Cheney said that seeing "new" usually means that something fishy is going on (in the conclusion).

Of course he is not The Absolute Source Of Truth, but I think that we should have an agreement about that.

zipper/zipper_test.go
gunnihinn: I do agree that we should think of ways to merge timeseries with different step time, but I disagree that we were doing that correctly before.

The commit I made reverts the situation back to before commit [f5a4fde940](https://github.com/go-graphite/carbonapi/commit/f5a4fde940#diff-0148271788c61e73bbce09e6a68e28c5L254), where we didn't attempt to merge metrics in that case. That commit introduced the new behavior, which has a bug:

In the merging function in current master, we take the metric of lower resolution and upsample its points to the resolution of the other metric. We do this by taking each bucket in the lower-resolution metric and expanding it to buckets of the size of the higher resolution metric. An example is helpful: Say `m1` has secondly resolution and we have 60 seconds of data, and `m2` has minutely resolution and we have 1 minute of data. We can represent these as two arrays of floats `m1 = [ a_1, ..., a_60 ]` and `m2 = [ b_1 ]`. We then expand `m2` to the array `m2' = [ c_1, ..., c_60 ]`, where `c_j = b_1` for all `j`.

For this to be correct, the aggregation function of the metric cannot be 'sum'. If it is, we have now corrupted the user's data, as `m2'` no longer downsamples to `m2`.

I think the right thing to do in this case is to revert to the previous behavior, which didn't try to deal with this case, but did the right thing. We should then open an issue and discuss how to implement merging of metrics with different time steps. As I wrote in the commit message, we have some thoughts on that over here and are not convinced this is as trivial as it might seem at first.

azhiltsov: We are talking about zipper here, right? And the sole purpose of zipper was to actually heal metrics from corruptions. Upscale metric will not have any gaps, so it will be always healing from upscaled metric and will create artificial data points. Reverting to original behavior is right thing to do here. We can discuss how to deal with such cases later if you feeling that there is such.

Civil:
TLDR; Discussed in Telegram chat and we agreed to merge it as is to make it working and then do another PR that will fix all issues mentioned here.

PR #346, opened 2018-07-21
Initial implementation of feature flag pkg

Civil:
There are more and more experimental changes. This pkg implements
a unified way to control what feature flags (experiment flags) enabled
and provides a way to change some of them in runtime (PATCH request).

From now on all code that considered to be experimental or not desirable by
everyone (e.x. incompatible with graphite-web) should be under a feature flag

Overhead of the implementation is low for single thread - 50ns per request,
Though it's better to replace whole function if the code path is too hot.

### DIFF
--- a/cmd/carbonapi/carbonapi.example.toml
+++ b/cmd/carbonapi/carbonapi.example.toml
@@ -40,6 +40,11 @@ file = "carbonapi.log"
 level = "info"
 logger = ""
 
+# features allows to specify feature flags on a startup time
+#   format: flag-name = value
+[featureFlags]
+applyByNode-count-nodes-from-1 = true
+
 # Uncomment this to get the behavior of graphite-web as proposed in https://github.com/graphite-project/graphite-web/pull/2239
 # Beware this will make darkbackground graphs less readable
 #[defaultColors]
--- a/cmd/carbonapi/carbonapi.example.yaml
+++ b/cmd/carbonapi/carbonapi.example.yaml
@@ -163,6 +163,10 @@ upstreams:
 # Default: 600 (10 minutes)
 graphTemplates: graphTemplates.example.yaml
 expireDelaySec: 10
+# features allows to specify feature flags on a startup time
+#   format: flag-name: value
+featureFlags:
+  applyByNode-count-nodes-from-1: true
 # Uncomment this to get the behavior of graphite-web as proposed in https://github.com/graphite-project/graphite-web/pull/2239
 # Beware this will make darkbackground graphs less readable
 #defaultColors:
--- a/cmd/carbonapi/config/config.go
+++ b/cmd/carbonapi/config/config.go
@@ -60,6 +60,7 @@ type ConfigType struct {
 	DefaultColors              map[string]string  `mapstructure:"defaultColors"`
 	GraphTemplates             string             `mapstructure:"graphTemplates"`
 	FunctionsConfigs           map[string]string  `mapstructure:"functionsConfig"`
+	FeatureFlags               map[string]bool    `mapstructure:"featureFlags"`
 
 	QueryCache cache.BytesCache `mapstructure:"-" json:"-"`
 	FindCache  cache.BytesCache `mapstructure:"-" json:"-"`
--- a/cmd/carbonapi/config/init.go
+++ b/cmd/carbonapi/config/init.go
@@ -17,6 +17,7 @@ import (
 	"github.com/go-graphite/carbonapi/expr/helper"
 	"github.com/go-graphite/carbonapi/expr/rewrite"
 	"github.com/go-graphite/carbonapi/limiter"
+	"github.com/go-graphite/carbonapi/pkg/features"
 	"github.com/go-graphite/carbonapi/pkg/parser"
 	zipperTypes "github.com/go-graphite/carbonapi/zipper/types"
 	"github.com/lomik/zapwriter"
@@ -52,6 +53,18 @@ func SetUpConfig(logger *zap.Logger, BuildVersion string) {
 			zap.Any("configuration", Config.Logger),
 			zap.Error(err),
 		)
+	}
+
+	flags := features.GetFeaturesInstance()
+	for flag, value := range Config.FeatureFlags {
+		err := flags.SetFlagByNameForced(flag, value)
+		if err != nil {
+			logger.Warn("failed to set feature flag",
+				zap.String("flag", flag),
+				zap.Bool("value", value),
+				zap.Error(err),
+			)
+		}
 	}
 
 	if Config.GraphTemplates != "" {
--- a/cmd/carbonapi/http/init.go
+++ b/cmd/carbonapi/http/init.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/dgryski/httputil"
+	"github.com/go-graphite/carbonapi/pkg/features"
 	"github.com/go-graphite/carbonapi/util/ctx"
 )
 
@@ -28,6 +29,11 @@ func InitHandlers() *http.ServeMux {
 
 	r.HandleFunc("/tags", tagHandler)
 	r.HandleFunc("/tags/", tagHandler)
+
+	features := features.GetFeaturesInstance()
+	r.HandleFunc("/_internal/flags", features.FlagListHandler)
+	r.HandleFunc("/_internal/flags/id", features.FlagPatchByIDHandler)
+	r.HandleFunc("/_internal/flags/name", features.FlagPatchByNameHandler)
 
 	r.HandleFunc("/", usageHandler)
 	return r
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -19,9 +19,20 @@ func (eval evaluator) EvalExpr(e parser.Expr, from, until int64, values map[pars
 
 var _evaluator = evaluator{}
 
+type rewriter struct{}
+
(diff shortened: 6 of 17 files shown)


gunnihinn:
Thank you for taking the time to propose an implementation of feature flags in the project. It's something we've discussed before, and all agree is a good thing to have.

If you put me on the spot and told me to implement a feature flag... err...  feature, I'd probably write

    type Features struct {
        foo bool
        bar bool
        baz someEnumType
        // other things that are value types, not pointers
    }

then initialize it from the config in my main package, and pass it to the HTTP handlers by value. To me, that implementation has some nice properties:

**(A)** It is thread safe by construction, and has the important property that once we start serving a request with a given feature setup, it is guaranteed to not change while we serve that request. We need that guarantee to be internally consistent in each request we serve. This is a must-have for any implementation, and is not fulfilled by this package, as its syncronization procedures live at too low a level to be able to help. We can also trivially implement HTTP handlers that modify the feature set at runtime, in an automatically thread-safe way.

**(B)** It is about as simple as one can imagine, and is obviously correct instead of having no obvious errors. That helps eliminate it as a source of problems when debugging in production.

**(C)** Having all feature flags explicitly named in one place servers as a kind of documentation about what features exist. It helps with the long-term maintenance of the project, as having a list of feature names we can grep for can quickly reveal dead code and unused features. It does mean we can't dynamically add new features at runtime, but since a feature needs a code deploy to do anything in the first place, this isn't a concern.

**(D)** Quote:

> Overhead of the implementation is low for single thread - 50ns per request, Though it's better to replace whole function if the code path is too hot.

That is the order of magnitude of performance your implementation will have because of its mutexes; see [Latency numbers every programmer should know](https://gist.github.com/jboner/2841832) for reference. In contrast, the dumb struct approach has as good performance as we can hope for without any further work.

I look forward to your responses to these comments, and your take on the dumb struct idea.

Civil:
>> If you put me on the spot and told me to implement a feature flag... err... feature, I'd probably write

This was the previous idea, yes. You can still see some traces of it in the code, you can have a look for traces of https://github.com/go-graphite/carbonapi/blob/69b6af58a484d0e694710ee55de61ac04e0e1317/cmd/carbonapi/main.go#L172. It wasn't struct though.

>> It is thread safe by construction, and has the important property that once we start serving a request with a given feature setup, it is guaranteed to not change while we serve that request. We need that guarantee to be internally consistent in each request we serve.

The same thing can be achieve if I'll take a snapshot of current struct in the beginning of the request and pass it around with Context. Basically in the beginning of every request we'll do atomic.Load for the whole feature set and use it during the whole lifetime. And for changing - atomic.Load -> change -> atomic.Store. I thought I've briefly mentioned that in telegram message. I just thought that for PoC it's less scary to have a simple mutex based implementation. See about performance below.

>> *(B)* It is about as simple as one can imagine, and is obviously correct instead of having no obvious errors. That helps eliminate it as a source of problems when debugging in production.

That's why even PoC have that amount of tests (code:tests ~ 1:3 at this moment). 

>> *(C)* Having all feature flags explicitly named in one place servers as a kind of documentation about what features exist. It helps with the long-term maintenance of the project, as having a list of feature names we can grep for can quickly reveal dead code and unused features.

It's easy to write a script that will generate markdown table with list of features and files where they are used. I think it should even be rather easy to write a proper project-specific linter (with AST parsing) to generate specific warnings for unused features, though I might be underestimating it's complexity at this moment.

>> *(D)* That is the order of magnitude of performance your implementation will have because of its mutexes; see Latency numbers every programmer should know for reference. In contrast, the dumb struct approach has as good performance as we can hope for without any further work.

That's actually hard to guesstimate. In worst case it won't be in cache, so you'll go to the main memory for that, and that's 100ns at least which is order of magnitude slower than mutex. To be honest my guess that even mutexed version will never show up in the profiling as a bottleneck. Especially if instead of calling to check for every ops we'll use flag for the whole code path in performance critical paths (function, etc). Also see answer to the *(A)* as it's easy to make it lock-free and have a constant overhead of one extra atomic ops per request, which I think will be around 10ns per request.



The reasons behind this complex thingy are:
1. It should make *your* operational work easier - most of potentially breaking changes will be covered by disabled-by-default-flag that you'd be able to control during runtime, without need to wait for code deploy, etc. It's also rather easy to add an API that will allow you to whitelist specific users for specific features - so proper A/B testing. I think that at this moment carbonapi and carbonzipper are complex enough to actually benefit from proper Canary and A/B.
2. It's possible to convince other projects (e.x. go-carbon) to adopt a framework to control features, but it will be nearly impossible to make them use a package with static struct. And eventually we'll end up in having feature flags in other packages and gently asking their maintainers to support getting those flags from us and behave accordingly.
3. It's much easier for separate developer to use init functions, this will encourage other people to use feature flags.
4. By it's definition feature flags are bool-only, so we need somehow enforce them to be like that. Easiest way to do that - on API level.
5. Having this as a sort of framework will enable us to have a centralized system that will control feature flags in future.

gunnihinn:
> The same thing can be achieve if I'll take a snapshot of current struct in the beginning of the request and pass it around with Context. Basically in the beginning of every request we'll do atomic.Load for the whole feature set and use it during the whole lifetime. And for changing - atomic.Load -> change -> atomic.Store. I thought I've briefly mentioned that in telegram message. I just thought that for PoC it's less scary to have a simple mutex based implementation.

We could do that. I'm not a huge fan of going through `interface{}` if I don't have to so the type system can help me out. Here it doesn't seem like we need to, we could have a main struct with the features, and wrap it in a struct

    type FeatureStore struct {
        f *Features      // our other struct
        m *sync.Mutex    // to update safely
    }

that we keep in our `main` function. When we start handling a request, we make a copy

    store.m.Lock()
    feats := *store.f
    store.m.Unlock()

and pass the copy down the request-handling funnel.

> That's why even PoC have that amount of tests (code:tests ~ 1:3 at this moment).

Writing tests is generally good, but a simpler implementation wouldn't need exhaustive tests. There's only so much that can go wrong with accessing struct fields. Tests also have a hard time proving the really important property we need; that features remain constant during a request.

> It's easy to write a script that will generate markdown table with list of features and files where they are used. I think it should even be rather easy to write a proper project-specific linter (with AST parsing) to generate specific warnings for unused features, though I might be underestimating it's complexity at this moment.

I suppose. However, given the choice between doing that work, and grepping for struct member names that are all in one place, I'll take the latter. It's not as exciting, but it also works 100% from the moment we write the struct definition, and requires no development or bugfixes to keep working forever.

> 2. It's possible to convince other projects (e.x. go-carbon) to adopt a framework to control features, but it will be nearly impossible to make them use a package with static struct. And eventually we'll end up in having feature flags in other packages and gently asking their maintainers to support getting those flags from us and behave accordingly.

I don't think we'll ever convince other projects to adopt a framework for features. Features are basically just parsed flags or config values, and as such are very project-specific. Getting our feature flags supported by other projects is also something I don't see a need for. If whatever we have is relevant to other projects, we can hopefully convince them of that on its own merits.

> 4. By it's definition feature flags are bool-only, so we need somehow enforce them to be like that.

That's not clear to me. I don't see why they couldn't be a value from a handful of options. If so, expressing the possibilities as combinations of boolean values gets painful fast, so any map implementation of the feature collection would likely become `map[string]interface{}`, and there goes type safety again.

Backing this by a map of any kind instead of a struct also makes things that could have been compile-time problems, like a misspelled feature name, into run-time problems. I really prefer if the compiler or type system can help me out in those cases.

Civil:
> We could do that. I'm not a huge fan of going through `interface{}` if I don't have to so the type system can help me out.

Sorry, I'm missing a point about `interface{}` here and I can't see where it comes from.

> Tests also have a hard time proving the really important property we need; that features remain constant during a request.

It can be easily proven - grab a snapshot, change the feature flag, grab another snapshot, compare current state and snapshots. If you see "Old, New, New" - it's fine. Also you can kinda rely on atomic.Load and atomic.Store to some extent.

In any case there must be a test that snapshoting actually works if the flags are runtime changeable, no matter if it's a struct or map[string]bool.

> I suppose. However, given the choice between doing that work, and grepping for struct member names that are all in one place, I'll take the latter. It's not as exciting, but it also works 100% from the moment we write the struct definition, and requires no development or bugfixes to keep working forever.

there is no difference between `grep SomeFeatureFlag *` and `grep checkFeature("flag-name") *`. And it's not true about bugfixes. Bugs can be anywhere, you can by mistake check for another feature or what so ever. It won't be a compile-time issue if you do a typo in a feature name, that's the only property you'll loose if we'll go for a `map[string]bool` direction instead of structs.

> That's not clear to me. I don't see why they couldn't be a value from a handful of options. If so, expressing the possibilities as combinations of boolean values gets painful fast, so any map implementation of the feature collection would likely become map[string]interface{}, and there goes type safety again.

By definition feature flags are booleans. I'm not talking here about config file where you can fine-tune stuff and play with the parameters. It's just a flags that you can change in a runtime (mostly) to choose one codepath (experimental) or another (stable). And nothing more than that. If we'll decide to implement proper A/B testing framework (where you can choose the percentage of requests affected by experimental code path or specific user name) it still won't change the flag layout to map[string]interface{}, as all those properties might be implemented in a structured way.

And to be honest I'll be strongly against non-bool feature flags in any form.

> and pass the copy down the request-handling funnel.
But this is what the context is about. It's by definition - request-scoped values, where you can access them whenever it's needed (if it's needed).

Civil:
Closing as infeasible at this moment.

PR #348, opened 2018-07-24
Gmagnusson/bugfix monitor warn

gunnihinn:
These are three unrelated patches that:

- Log a new warning when Find requests in Fetch fail
- Expose the carbonzipper request duration buckets in carbonapi for monitoring
- Fix an infinite loop bug in Fetch requests that happens when we filter out some clients

### DIFF
--- a/cmd/carbonapi/main.go
+++ b/cmd/carbonapi/main.go
@@ -12,6 +12,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 	"unicode"
 
@@ -43,6 +44,7 @@ var apiMetrics = struct {
 	RequestCacheHits      *expvar.Int
 	RequestCacheMisses    *expvar.Int
 	RenderCacheOverheadNS *expvar.Int
+	RequestBuckets        expvar.Func
 
 	FindRequests        *expvar.Int
 	FindCacheHits       *expvar.Int
@@ -172,6 +174,7 @@ var config = struct {
 	ExtrapolateExperiment      bool               `mapstructure:"extrapolateExperiment"`
 	Logger                     []zapwriter.Config `mapstructure:"logger"`
 	Listen                     string             `mapstructure:"listen"`
+	Buckets                    int                `mapstructure:"buckets"`
 	Concurency                 int                `mapstructure:"concurency"`
 	Cache                      cacheConfig        `mapstructure:"cache"`
 	Cpus                       int                `mapstructure:"cpus"`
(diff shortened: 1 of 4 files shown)


cmd/carbonapi/main.go
```diff
@@ -496,6 +504,10 @@ func setUpConfig(logger *zap.Logger) {
 		graphite.Register(fmt.Sprintf("%s.request_cache_misses", pattern), apiMetrics.RequestCacheMisses)
 		graphite.Register(fmt.Sprintf("%s.request_cache_overhead_ns", pattern), apiMetrics.RenderCacheOverheadNS)
 
+		for i := 0; i <= config.Buckets; i++ {
+			graphite.Register(fmt.Sprintf("%s.requests_in_%dms_to_%dms", pattern, i*100, (i+1)*100), bucketEntry(i))
```
Civil: There was an old Feature-request for carbonzipper and carbonapi to call last bucket as `requests_in_Xms_to_inf`. Maybe as it's a good idea to implement that as part of this PR, otherwise we'll end up having `(i+1)*100` for last bucket forever.

zipper/broadcast/broadcast_group.go
```diff
@@ -192,6 +197,8 @@ func (bg *BroadcastGroup) Fetch(ctx context.Context, request *protov3.MultiFetch
 	}
 
 	requests := bg.SplitRequest(ctx, request)
+	// TODO(gmagnusson): WAIT, HOW MANY METRICS WAS THAT
```
Civil: Btw, zap supports sampling of the logs - you can specify how many messages per second (max) for particular logger you want to have.

It might be a good idea for future improvement to implement an HTTP Handle that allows you to temporary bump loglevel and to control sampling rate.

cmd/carbonapi/main.go
Civil: See:
https://godoc.org/go.uber.org/zap#SamplingConfig
https://godoc.org/go.uber.org/zap#AtomicLevel.ServeHTTP

and docs around that.

gunnihinn: We thought about it and decided we'd like the last bucket to also be `requests_in_Xms_to_(X+100)ms`, because if we don't and change our minds at some point about how many buckets we want to have (say we'd like to have 20 instead of 10), then we'll have two `_to_inf` buckets lying around in Graphite.

Civil: But in that case the first "to_inf" bucket will contain only nulls after that change. Otherwise it'll continue to exist, but the meaning of the values will suddenly change.

I understand that it's actually choosing lesser of two evils, so up to you to make decision on default behavior. If you think it's ok, feel free to merge PR.

I'd still suggest to at least add an option in future that will allow to change that behavior.

gunnihinn: Both make sense. I vote we defer this to [an issue](https://github.com/go-graphite/carbonapi/issues/350) so we can get the non-infinite-loop fix in here. For now, the situation will at least be no worse than in carbonzipper.

Civil: Agree.

PR #354, opened 2018-07-30
changing default golang we use for packaging to 1.10.x instead of beta

azhiltsov:
Periodically travis fails to build packages with errors like:
# github.com/go-graphite/carbonapi/vendor/go.uber.org/zap
vendor/go.uber.org/zap/field.go:33: syntax error: unexpected = in type declaration

one of the reasons I can imagine for that is this:
go version go1.11beta2 linux/amd64

additionally according to https://docs.travis-ci.com/user/languages/go/
changing go version declaration to string instead of float.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,9 @@ git:
 
 matrix:
   include:
-    - go: 1.x
-      env:
+    - go: "1.x"
+    - go: "1.9.x"
+    - go: "1.10.x"
+       env:
         - BUILD_PACKAGES=true
-    - go: 1.9.x
-    - go: 1.10.x
-    - go: master
+   - go: "master"


Civil:
Could you please try to update vendored dep?

Reason why we are testing beta golabg there is to catch those errors earlier than later. So there is now a chance that this thing will be fixed before go 1.11 release (no matter if it's golang or zap bug)

azhiltsov:
The point is: we shouldn't build and ship GA packages using beta golang.
That is why BUILD_PACKAGES=true in my PR is moved to 1.10.x which still able to compile
You can check any build, for instance https://travis-ci.org/go-graphite/carbonapi/builds/409753237
Travis will keep building using all specified go versions and we can discuss builds with beta and dep ensure in separate tread.

Civil:
Golang in Travis is not used to build packages.

Packages are built in Docker in provided images, using golang that's in those images.

PR #355, opened 2018-07-30
Remove unused vendored dependencies

gunnihinn:
Dep was complaining about them.


gunnihinn:
I can't merge this because the CI pipeline fails on the package building step.

Civil:
Hm, I thought you had an admin privileges on this repo. Now you have and you can force-merge this things.

Civil:
@azhiltsov sorry, but I disagree with you on both points. The whole purpose of testing with "1.x" and "master" is to catch regressions as early as possible, so removing them is not an option.
Not caring about compile breakage in a Dependency is also not an option as we are those who will live with it. Proper way to solve this would be reporting a bug upstream (though it seems that it's now fixed).

PR #357, opened 2018-07-31
Get some request statistics

gunnihinn:
This series of commits is aimed at the metrics we write to Graphite. We delete some dead code, populate the zipper.Stats we pass around in the various requests, and count carbonapi total requests and errors.

### DIFF
--- a/cmd/carbonapi/http_handlers.go
+++ b/cmd/carbonapi/http_handlers.go
@@ -152,6 +152,7 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
 
 	err := r.ParseForm()
 	if err != nil {
+		apiMetrics.Errors.Add(1)
 		http.Error(w, http.StatusText(http.StatusBadRequest)+": "+err.Error(), http.StatusBadRequest)
 		accessLogDetails.HTTPCode = http.StatusBadRequest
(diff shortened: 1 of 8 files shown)


Civil:
Purpose of the struct was to contain all required stats for the request. It might not necessary be exposed as a metric, but at some point in the past it was in the access logs for the request. Those indeed were broken since the whole zipper rewrite thing (but they were used in last stable release)

Stats.SearchRequests - was used to count requests to CarbonSearch.
Stats.Cache{Hits,Misses} - cache misses and hits.
Stats.Timeouts - timouts that happened inside the zipper.
Stats.MemoryUsage - it should contain some approximation about memory usage per request, we used to search logs with by this stat to filter out heavy requests.

To be honest I don't think that removing that and adding it back in a separate commit is a great idea, as we actually should populate those stats where necessary.

gunnihinn:
> To be honest I don't think that removing that and adding it back in a separate commit is a great idea, as we actually should populate and expose those stats again, unless you have a good arguments why they are not useful.

I'm fine with exposing more metrics, I just didn't have any context for what these were supposed to do and didn't want to keep code that wasn't doing anything around. I can rework the PR to make these do things again, unless you add some commits on top of these. :)

PR #364, opened 2018-09-26
pixelRatio param for render with format=png

lomik:
Optional `pixelRatio` param was added. It is usable for drawing more clear images for retina (or 4k) displays

Example result on 4k display (left image with pixelRatio=2.0, right with default): 
screen shot 2018-09-26 at 23 42 39


Civil:
I'm not sure why, but travis-ci decided to ignore this PR and I found no way how to refresh it.

PR #384, opened 2019-01-29
Add seriesByTag and aliasByTags

kolobaev:
Add functions, for graphite-tags support.


Civil:
It still would benefit from support on carbonzipper side, however I think that would be a good start.

Issue #403, opened 2019-07-10
Bad rendering using fallback graphiteWeb 


sylvain-beugin:
With latest version 0.12.2 ,when using fallback to graphiteWeb, carbonapi do the request to graphiteWeb correctly (even with maxDatapoint option) but return bad points in response (value is the same as date)

[{"target":"smartSummarize(server.collectd.cpu-0.cpu-idle, \"10min\", \"max\")","datapoints":[[**1562584320**,**1562584320**],[1562584920,1562584920],[1562585520,1562585520],....]}]

I have patched expr/functions/graphiteWeb/function.go  and it seems to correct this bug:

 for i, v := range m.Datapoints {
                        **pbResp.Values[i] = v[0]**
                }

I use graphite web 1.1.4 and value is always in first element:

 curl "http://localhost/graphite/render/?format=json&from=1562733833&target=smartSummarize%28mservercollectd.cpu-0.cpu-idle%2C+%2710min%27%2C+%27max%27%29&until=1562755433"

[{"datapoints": [[**99.033587**, **1562733840**], [98.965809, 1562734440],

Civil:
Yeah, that might be the case. I thought I've corrected that issue.

Could you send a PR to fix that?

Civil:
Merged it. Thanks.
P.S. If you specify in the commit `Fixes #403` github will automatically close the issue :)

PR #404, opened 2019-07-10
Correct  #403 bad parsing of datapoints

sylvain-beugin:
value is always in first element , date in the the second element  of datapoint json array deleiver by graphite web


sylvain-beugin:
see issue https://github.com/go-graphite/carbonapi/pull/404

Civil:
Merged it. Thanks.

PR #442, opened 2020-01-14
add ignoreHTTPError option

faceair:
In graphite-web, all rendering errors will only return empty results, no errors will be returned. For example, divide by an empty series.

It may not be reasonable, but I want them to be consistent, we have some systems that expect the graphite api to never return an error.


Civil:
I agree that consistency is important (and even should be default behavior whenever it is sane to do it), however PR currently changes the behavior too much.

However I'll need a bit more time to test the PR.

faceair:
I noticed that [this commit ](https://github.com/go-graphite/carbonapi/commit/3ae9ab7815bd5124d2a0fc4f533f9d2159cd9a36
) can solve this problem, this mr does not need to be merged.

PR #444, opened 2020-01-23
sortBy implementation

misiek08:
Implementing `sortBy/3` to match graphite's ones. I will try to compare results from Graphite and my changes (this and https://github.com/go-graphite/carbonapi/pull/437) and if there are differences I'll try to fix them in following PRs.

No tests changes for existing `sortBy*` functions, so no breaking changes for this field.

I'm open for some optimizations proposals. I wasn't trying to make it as fast as possible here by complicating code, but there is place for speeding up this code if needed.


misiek08:
@Civil thanks 😉

PR #456, opened 2020-03-27
fix applyByNode

faceair:
When querying `aliasByNode(highestAverage(applyByNode(host.dns[0-9]*.uptime.uptime.value,1,"scale(perSecond(%.interface-bond1.if_octets.tx), 0.000008)"), 5), 1, 2)`, the `function = aliasByNode, err = function = highestAverage, err = unknown function in evalExpr: "applyByNode"` error still occurs.

Looks like the rewrite function cannot work with other functions, I try to implement applyByNode in other ways to avoid this error.

### DIFF
--- a/cmd/carbonapi/http/helper.go
+++ b/cmd/carbonapi/http/helper.go
@@ -2,7 +2,6 @@ package http
 
 import (
 	"fmt"
-	"github.com/go-graphite/carbonapi/pkg/parser"
 	"net/http"
 	"strings"
 	"sync/atomic"
@@ -10,6 +9,7 @@ import (
 
 	"github.com/go-graphite/carbonapi/carbonapipb"
 	"github.com/go-graphite/carbonapi/cmd/carbonapi/config"
+	"github.com/go-graphite/carbonapi/pkg/parser"
 	"github.com/lomik/zapwriter"
 	"go.uber.org/zap"
 )
@@ -18,11 +18,6 @@ type responseFormat int
 
 // for testing
 var timeNow = time.Now
-
-type requestInterval struct {
-	from  int64
-	until int64
-}
 
 const (
 	jsonFormat responseFormat = iota
--- a/cmd/carbonapi/http/render_handler.go
+++ b/cmd/carbonapi/http/render_handler.go
@@ -1,13 +1,13 @@
 package http
 
 import (
-	"github.com/ansel1/merry"
 	"io/ioutil"
 	"net/http"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/ansel1/merry"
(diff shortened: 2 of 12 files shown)


Civil:
Could you also please modify the test for applyByNode function? As they will fail because of the node index change.

faceair:
I have supported mixed query of rewrite function and other functions, but it seems that I cannot add some details to the access log. Do you have any good suggestions ?

Civil:
What kind of details are you trying to add?

faceair:
such as:
1. https://github.com/go-graphite/carbonapi/pull/456/files#diff-fa3458f1b1f03af74514bd42a0fc2ec4L274-L275
2. https://github.com/go-graphite/carbonapi/pull/456/files#diff-fa3458f1b1f03af74514bd42a0fc2ec4L319

Civil:
Not sure that I understood your question correctly. I also haven't reviewed your latest changes (yet).

But overall accessLogDetails is a struct (https://github.com/go-graphite/carbonapi/blob/master/carbonapipb/carbonapi.pb.go#L3-L34) that supposed to be populated whenever you have something to add there.

Idea behind it was that in each HTTP handler there should be a defer that will produce a log entry in access log with all the details from that struct. For example that one: https://github.com/go-graphite/carbonapi/blob/master/cmd/carbonapi/http/render_handler.go#L97-L99 for `/render`.

defferedAccessLogger is the function defined in helper and just sets the HTTP response code if needed and specify text itself:
https://github.com/go-graphite/carbonapi/blob/b99b96f80a8302ac29e9ee60bd6f7f626694e650/cmd/carbonapi/http/helper.go#L241-L249

As about the stats:
They come from zipper itself and should be populated by respective backend.

For render requests that usually populated in the Fetch function in the broadcast group:
https://github.com/go-graphite/carbonapi/blob/8dad9e0b5dcda1481d84442728c95e2e2df51d85/zipper/broadcast/broadcast_group.go#L244

And the function itself:
https://github.com/go-graphite/carbonapi/blob/8dad9e0b5dcda1481d84442728c95e2e2df51d85/zipper/broadcast/broadcast_group.go#L302-L313

Also ZipperRequests here have a different meaning than in carbonapi 0.11 or earlier (or booking's fork as it's based on 0.11) - as zipper now is integrated into carbonapi, this metric have a meaning of how many backend functions we need to call after zipper split our requests according to the configured parameters.

But actually I can see your point that it's currently not the real amount of requests, but the prediction. Those metrics should be populated at least partially after that point:
https://github.com/go-graphite/carbonapi/blob/8dad9e0b5dcda1481d84442728c95e2e2df51d85/zipper/broadcast/broadcast_group.go#L258

Civil:
Thanks for the PR, it's really helpful

PR #459, opened 2020-04-02
ignore no metrics error

faceair:
1. Fix the following case, which is caused by the rewrite function fetch.

query:
```
aliasByNode(highestAverage(applyByNode(host.dns[0-9]*.uptime.uptime.value,1, "scale(perSecond(%.interface-bond1.if_octets.tx), 0.000008)"), 5), 1, 2)
```
error:
```
Internal Server Error: error or no response: function=aliasByNode, err=function=highestAverage, err=no metrics in the Response
```

2. Another commit fixes same metric name cache

for example:
```
diffSeries(flink.1165.jobmanager.*.*.numberOfFailedCheckpoints, timeShift(flink.1165.jobmanager.*.*.numberOfFailedCheckpoints, '10min'))
```

### DIFF
--- a/cmd/carbonapi/http/render_handler.go
+++ b/cmd/carbonapi/http/render_handler.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-graphite/carbonapi/expr/types"
 	"github.com/go-graphite/carbonapi/pkg/parser"
 	utilctx "github.com/go-graphite/carbonapi/util/ctx"
+	ztypes "github.com/go-graphite/carbonapi/zipper/types"
 	pb "github.com/go-graphite/protocol/carbonapi_v3_pb"
 	"github.com/lomik/zapwriter"
 	uuid "github.com/satori/go.uuid"
@@ -270,6 +271,9 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
 		// Otherwise it should be 500
 		errMsgs := make([]string, 0)
 		for _, err := range errors {
+			if merry.Is(err, ztypes.ErrNoMetricsFetched) {
+				continue
+			}
 			errMsgs = append(errMsgs, err.Error())
 			if returnCode < 500 {
 				returnCode = merry.HTTPCode(err)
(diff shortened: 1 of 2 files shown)


cmd/carbonapi/http/render_handler.go
```diff
@@ -260,6 +261,9 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
 		// Otherwise it should be 500
 		errMsgs := make([]string, 0)
 		for _, err := range errors {
+			if err != ztypes.ErrNoMetricsFetched {
```
Civil, line 274: It's better to use `!merry.Is(err, ztypes.ErrNoMetricsFetched)` here as `!=` will only compare one level of the errors, but "Is" will check if on any level there is an error of that type. That's not too important, but would allow to change all types of the errors much easier in future.

Also I'm a bit confused with the PR's name - you are saying that you want to ignore "no metircs error", but here you are ignoring everything, but "NoMetircsFetched".

faceair: Yes, it should be `merry.Is(err, ztypes.ErrNoMetricsFetched)`

faceair: image

But it looks like this error is not equal to ErrNoMetricsFetched, it is equal to https://github.com/go-graphite/carbonapi/blob/c538fcd86be320b360e62fd0a6dd7d136b428e87/expr/expr.go#L129

We may need to modify the error handling here.

faceair:
By the way, I don't think we need to generate an ErrNoMetricsFetched error by default when backend fetch, just return an empty response directly. 
If a function requires the value during processing, this time the function can throw an ErrNoMetricsFetched error.

Civil:
Zipper currently generates ErrNoMetricsFetched because it tries to stay somehow decoupled from carbonapi. Even though it's not very well maintained, but there is cmd/carbonzipper app that can be used separately from carbonapi, at least for now.

PR #464, opened 2020-05-06
Fix /render empty result returns 200.

lexx-bright:


cmd/carbonapi/http/render_handler.go
```diff
@@ -269,15 +269,14 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
...
 		errMsgs := make([]string, 0)
 		for _, err := range errors {
 			if merry.Is(err, ztypes.ErrNoMetricsFetched) || merry.Is(err, parser.ErrSeriesDoesNotExist) {
 				continue
 			}
 			errMsgs = append(errMsgs, err.Error())
-			if returnCode < 500 {
-				returnCode = merry.HTTPCode(err)
-			}
+			returnCode = merry.HTTPCode(err)
```
Civil, line 279: I've explicitly wanted to any error that would cause 5xx error to return 5xx, e.x. you have 2 backends that reported 404, no metrics found, and one that returned "500" due to some error should cause overall return code to be 500.

In that case it will strictly depend on order of the replies and 404 could override 500.

cmd/carbonapi/http/render_handler.go
```diff
@@ -269,15 +269,14 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
 		// Obtain error code from the errors
 		// In case we have only "Not Found" errors, result should be 404
 		// Otherwise it should be 500
+                returnCode = http.StatusNotFound
```
Civil, line 272: Please run `go fmt` on the file to correct formating.

cmd/carbonapi/http/render_handler.go
lexx-bright: Refactored to follow that logic.

cmd/carbonapi/http/render_handler.go
lexx-bright: Done.

lexx-bright:
Got the idea. Predefining returnCode with http.StatusNotFound in case of empty result should fix this issue.

PR #473, opened 2020-05-28
add start/end time to the metric find request

faceair:


cmd/carbonapi/interfaces/zipper.go
```diff
@@ -12,7 +13,7 @@ import (
 // The CarbonZipper interface exposes access to realZipper
 // Exposes the functionality to find, get info or render metrics.
 type CarbonZipper interface {
-	Find(ctx context.Context, metrics []string) (*pb.MultiGlobResponse, *zipperTypes.Stats, merry.Error)
+	Find(ctx context.Context, request pb.MultiGlobRequest) (*pb.MultiGlobResponse, *zipperTypes.Stats, merry.Error)
```
faceair, line 16: Should I use struct or struct pointer here? Are there any principles?

Civil: In general, it doesn't matter if the struct is small enough. Basically if it's less than 60 bytes that's fine either way, otherwise pointer is better to avoid extra copy of "large" structure.

But as this struct only contains couple more int64 fields that should be fine as a struct. Also other functions uses struct at this moment, so either all of them should use pointers or none of them, just to preserve same semantics across all of the functions.

faceair: Thanks, I get it.

Issue #18, opened 2014-09-27
Cache more aggressively


dgryski:
Graphite munges timestamps to minutely resolution to increase the chance of a cache hit. It also stores the data returned by the stores so future queries on the same data can also be pulled from the cache.

Decide which of these we want to implement.

xneo64:
@dgryski why not make `cacheTimeout := int32(60)` duration user configurable? 

Graphite-web implements DEFAULT_CACHE_DURATION which is very useful in my use case as we collect metrics at an interval of 10s. 
When using carbonapi I have to manually override the Cache Timeout value in Grafana, or disable caching altogether - both of which are either impractical or undesirable.

dgryski:
@xneo64 Good plan.  I'll do that on Monday when I'm back at work.  (If you have a PR, I can review and merge that sooner.)

xneo64:
Just got to know about this project a couple of days ago, and converted my preproduction environments. While very easy to read, unfortunately I'm inexperienced with Go code, so I'll do you a favour by not submitting a barbaric PR. 

 I'll definitely be able to make proper + tested contributions in the future

dgryski:
The munged timestamps to increase cache hit ratio also an approach that should be considered. Would be interesting to see what effect it has on real traffic. We can probably simulate this given logs.

jaroslawr:
@dgryski @Civil are there any realistic near-term plans on implementing the timestamp munging? If you consider handling requests from Grafana, right now caching only has a chance to go in effect if two or more identical queries arive within a single second, right? To me it looks like operating with full-minute timestamps would have a night-and-day performance effect for many production set ups. I am troubleshooting one right now, and the bulk of carbonapi requests are popular Grafana dashboards which repeatedly issue the same queries with request time ranges like [-24h,now] etc., which unfortunately do not get cached because the timestamps are resolved differently even if sent 5s or 10s apart.

Civil:
I'm open to a PR or at least to some one who are willing to test such a feature. As the overall problem here is that neither me nor Damian have production environments that uses carbonapi (at least to my knowledge), so it is very hard to test what benefit it would have in real life and if it won't add any new bugs (bugs within caching could be tricky to diagnose).

Civil:
Actually there was another idea on how to improve loading time for dashboards and hit ratio as well:
https://github.com/go-graphite/carbonapi/issues/242

But those approaches can co-exist.

jaroslawr:
[...]

PR #480, opened 2020-06-16
Add a new cache for caching backend responses before rendering.

jaroslawr:
The new cache has a much better hit ratio when using CarbonAPI with Grafana. Grafana varies the value of maxDataPoints sent in requests depending on the screen width of the client, which often prevents reusing the results in the existing cache, which stores the response already after trimming the results to maxDataPoints. See also https://github.com/go-graphite/carbonapi/issues/18

The new cache is free from this problem and for us it results in around 25-30% less queries to Clickhouse and 25%-30% less load on Clickhouse servers, which we use for storing the metrics, in comparison to a setup with just the existing  cache.

I tried to be fairly conservative in implementing this:
- the preexisting response cache is still there
- the backend cache is configured separately and is disabled by default
- noCache param disables both caches (and cacheTimeout param overwrites the timeout for both caches)
- if there are errors when fetching the metrics from backends, nothing is stored in the cache

### DIFF
--- a/cmd/carbonapi/http/render_handler.go
+++ b/cmd/carbonapi/http/render_handler.go
@@ -1,6 +1,9 @@
 package http
 
 import (
+	"bytes"
+	"encoding/gob"
+	"errors"
 	"io/ioutil"
 	"net/http"
 	"strconv"
@@ -41,9 +44,7 @@ func setError(w http.ResponseWriter, accessLogDetails *carbonapipb.AccessLogDeta
 	accessLogDetails.HTTPCode = int32(status)
 }
 
-func getCacheTimeout(logger *zap.Logger, r *http.Request) int32 {
-	cacheTimeout := config.Config.Cache.DefaultTimeoutSec
-
+func getCacheTimeout(logger *zap.Logger, r *http.Request, defaultTimeout int32) int32 {
 	if tstr := r.FormValue("cacheTimeout"); tstr != "" {
 		t, err := strconv.Atoi(tstr)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


cmd/carbonapi/http/render_handler.go
```diff
@@ -332,11 +348,62 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
...
+		config.Config.ResponseCache.Set(responseCacheKey, body, responseCacheTimeout)
 		td := time.Since(tc).Nanoseconds()
 		ApiMetrics.RenderCacheOverheadNS.Add(td)
 	}
 
 	gotErrors := len(errors) > 0
 	accessLogDetails.HaveNonFatalErrors = gotErrors
 }
+
+func backendCacheComputeKey(from string, until string, targets []string) string {
```
Civil, line 360: you can combine from and until there. I mean instead of
```
func backendCacheComputeKey(from string, until string, targets []string) string {
```

you can write
```
func backendCacheComputeKey(from, until string, targets []string) string {
```

That's just to make the linter a bit happier.

Otherwise PR looks good to me on a first glance.

Meanwhile I'll read it more careful.

jaroslawr: Thanks, I amended the commit with a fix.

jaroslawr:
@Civil My PR was still for the master branch, it seems main is the default branch now. There are also some other differences between main and master branch now. Is that somehow intentional, or just a result of the branch rename?

In any case, thanks for merging!

Civil:
I've fixed that. Maybe a bit suboptimal, but current main should have all the PRs that were mis-merged.

Sorry for a confusion. That wasn't intentional, I just assumed that github is smarter than it really is.

PR #129, opened 2016-06-30
[do not merge] WIP: sorting for globbed metrics

arodland:
I had some inspiration on https://github.com/dgryski/carbonapi/issues/128 and did some work. It's a ways from production quality but I'd appreciate any feedback.

PR #493, opened 2020-07-07
Sort results before answer request

Felixoid:
graphite-web always sort metric names in the response  
screen_2020-07-07_10-36-01_window

But carbonapi didn't and results were inconsistent  
screen_2020-07-07_10-34-48_window  
screen_2020-07-07_10-35-30_window

This PR implements sorting of final results before an answer.

### DIFF
--- a/cmd/carbonapi/http/render_handler.go
+++ b/cmd/carbonapi/http/render_handler.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"io/ioutil"
 	"net/http"
+	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


arodland:
CarbonAPI *had* this behavior — it was added in https://github.com/go-graphite/carbonapi/pull/129 from 2016. Somehow it got lost in between :)

Civil:
My theory that it was lost during refactoring as that doesn't happen with go-carbon (or wasn't matter for the most of the queries and no one complained after it happened).

PR #499, opened 2020-07-16
Add render panic HTTP answer

Felixoid:
Currently, carbonapi answers with 200 without any signs of errors except a render log.

This PR adds 500 status code and sending stack trace (configurable, enabled by default).

### DIFF
--- a/cmd/carbonapi/config/config.go
+++ b/cmd/carbonapi/config/config.go
@@ -78,6 +78,7 @@ type ConfigType struct {
 	Prefix                     string             `mapstructure:"prefix"`
 	Expvar                     ExpvarConfig       `mapstructure:"expvar"`
 	NotFoundStatusCode         int                `mapstructure:"notFoundStatusCode"`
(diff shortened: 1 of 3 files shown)


Civil:
Other than a small typo in docs I can't see any problems with PR (well, I'm not a big fan of doing more things within `recover`, but I can't see any better option.

expr/functions/asPercent/function.go
```diff
@@ -86,6 +86,7 @@ func (f *asPercent) Do(ctx context.Context, e parser.Expr, from, until int64, va
 			return nil, types.ErrWildcardNotAllowed
 		}
 		if len(total) == 1 {
+			total = helper.AlignSeries(append(total, arg...))[0:1]
```
Civil, line 89: I think this is a left-over from the other PR you've opened:
https://github.com/go-graphite/carbonapi/pull/500

Could you please remove that part from this PR?

Felixoid: Yep, wrong rebase. Thanks

PR #500, opened 2020-07-16
Align total and first series in asPercent function

Felixoid:
Currently, the issues like `runtime error: index out of range [288] with length 288` are possible. The stack-trace is

```
 0  0x000000000043aa64 in runtime.goPanicIndex
    at /usr/lib/go/src/runtime/panic.go:88
 1  0x0000000000f320c6 in github.com/go-graphite/carbonapi/expr/functions/asPercent.(*asPercent).Do.func5
    at ./expr/functions/asPercent/function.go:90
 2  0x0000000000f2ef17 in github.com/go-graphite/carbonapi/expr/functions/asPercent.(*asPercent).Do
    at ./expr/functions/asPercent/function.go:284
 3  0x0000000001004e05 in github.com/go-graphite/carbonapi/expr.EvalExpr
```

Here is a list of changes:

- Align total and first series in asPercent function
- Add tests helper for case of modified initial metrics
- Add a check for StepTime and adjust tests
- Add a check for StartTime and adjust tests
- Add a check for StopTime, adjust tests and fix functions

### DIFF
--- a/expr/functions/aggregateLine/function.go
+++ b/expr/functions/aggregateLine/function.go
@@ -93,6 +93,7 @@ func (f *aggregateLine) Do(ctx context.Context, e parser.Expr, from, until int64
 		} else {
 			r.FetchResponse.StepTime = a.FetchResponse.StopTime - a.FetchResponse.StartTime
 			r.FetchResponse.Values = []float64{val, val}
+			r.FetchResponse.StopTime = r.StartTime + int64(len(r.FetchResponse.Values))*r.FetchResponse.StepTime
 		}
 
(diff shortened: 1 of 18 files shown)


Civil:
Maybe it would be better to just unconditinionally align series as soon as you have them? E.x. before line 50 of the file. As I fear alignment would be required anyway in all possible cases. Otherwise there would be possible to cause panic anyway.

Felixoid:
I'm on it. Will try to find every place where it could be a problem. And I'll add tests to cover edge-cases

Felixoid:
ready for review

PR #501, opened 2020-07-20
Move precision alignment from per-target zipper to per-request FetchAndEvalExp

Felixoid:
It fixes an issue from #496.

With per-target strategy, it doesn't align StepTime always when it's needed. And sometimes the other way around, it aligns unnecessary values.

Changes here:

- Implement MetricData.Copy function
- Move ScaleToCommonStep logic from zipper to expr.FetchAndEvalExp
- Optimize GetCommonStep function
- Simplify powSeriesLists and make(map) in weightedAverage for DeepSource

### DIFF
--- a/expr/helper/align.go
+++ b/expr/helper/align.go
@@ -37,20 +37,19 @@ func LCM(args ...int64) int64 {
 // GetCommonStep returns LCM(steps), changed (bool) for slice of metrics.
 // If all metrics have the same step, changed == false.
 func GetCommonStep(args []*types.MetricData) (commonStep int64, changed bool) {
-	changed = false
-	steps := make([]int64, 0, len(args))
-	firstStep := args[0].StepTime
+	steps := make([]int64, 0, 1)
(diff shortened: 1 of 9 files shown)


expr/helper/align.go
```diff
@@ -33,25 +34,50 @@ func LCM(args ...int64) int64 {
...
-func GetCommonStep(args []*types.MetricData) int64 {
+// GetCommonStep returns LCM(steps), changed (bool) for slice of metrics.
+// If all metrics has the same step, changed == false.
+func GetCommonStep(args []*types.MetricData) (commonStep int64, changed bool) {
+	changed = false
 	steps := make([]int64, 0, len(args))
+	firstStep := args[0].StepTime
 	for _, arg := range args {
 		steps = append(steps, arg.StepTime)
+		if !changed && firstStep != arg.StepTime {
```
Felixoid, line 45: It makes sense to check if a step is already in steps. Will see, how to do it in faster way

PR #511, opened 2020-08-05
Recalculate metrics start/step/stop for backends

Felixoid:
Always recalculate metrics to the same [formula](https://github.com/go-graphite/carbonapi/blob/e7ac8954453cfcccfb0016ffc639977b8a325a30/expr/types/types.go#L393) for consistency.

Otherwise, helper.AlighSeries doesn't work as expected.

see lomik/graphite-clickhouse#84


zipper/broadcast/broadcast_group.go
```diff
@@ -288,9 +288,14 @@ func (bg *BroadcastGroup) Fetch(ctx context.Context, request *protov3.MultiFetch
...
-		} else {
-			return nil, result.Stats, types.ErrNotFound.WithHTTPCode(404)
 		}
+		return nil, result.Stats, types.ErrNotFound.WithHTTPCode(404)
+	}
+
+	// Recalculate metrics start/step/stop parameters to avoid upstream misbehavior
+	for i, metric := range result.Response.Metrics {
+		result.Response.Metrics[i].StopTime = metric.StartTime + int64(len(metric.Values))*metric.StepTime
+		_, _ = metric.StopTime, i
```
Civil, line 298: Not sure why this is needed?

Felixoid: sorry, it's for debugging.

PR #527, opened 2020-10-08
Fix groupByNode(s) name changing

Felixoid:
It fixes the issue reported by @lomik

I've checked it locally, looks good.


Civil (approved):

Civil:
I think I should rethink the way how name tag is handled.

PR #541, opened 2020-11-08
Backend-specific maxBatchSize should have priority over global settings

Civil:
Backend-specific maxBatchSize should have priority over global
    
    Current way of parsing (global have priority over specific) is confusing for people.
    
    Also current way of config parsing doesn't allow to enforce stricter or more relaxed rules depending on backend and
    doesn't work as it's intended, making per-backend setting useless.
    
    Rework request split logic to prefer backend-specific setting.


Felixoid (approved):
Tested it, works fine. AFAI understand, splitRequest used to use bg settings, and now it inherits settings from the backend properly :+1:

Issue #521, opened 2020-09-11
[FR] Use Graphite Metrics API instead of `/api/v1/series` when exploring metrics in VictoriaMetrics backend


valyala:
**Is your feature request related to a problem? Please describe.**

Currently carbonapi queries metrics via [/api/v1/series](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers) when using Prometheus backend. This may be slow when the backend contains millions of metrics that start with small number of distinct nodes, since it should fetch all the millions of metric names and then filter a few distinct nodes from them.

**Describe the solution you'd like**

Certain Prometheus-compatible backends such as VictoriaMetrics support Graphite Metrics API (see [these docs](https://victoriametrics.github.io/#graphite-metrics-api-usage)), which have much better performance when exploring big number of metric names. It would be great if carbonapi could use optional Graphite Metrics API when using Prometheus backend.

Civil:
I had a look at the provided API, but it seems that support of /metrics/find is not graphite-compatible (only format=completer and format=treejson are supported) which makes it harder to utilize.

Basically with graphite-web currently it's possible to do a query like that:
```
/metrics/find/?query=*.foo.{bar,baz}.bam&format=msgpack
```

And it would give exact list of metrics (same works with format=json)

However victoriametrics supports only treejson and completer. Both would give names that looks like:
`*.foo.{bar,baz}.bam` without specifying which of `bar` or `baz` matched and what `*` expands to.

So to construct the full list of metrics I would need to implement custom logic that would do a globbing on a carbonapi side which I would like to avoid.

valyala:
Full list of matching metrics can be obtained from `/metrics/expand/` handler. Try sending request to `/metrics/expand/?query=*.foo.{bar,baz}.bam`

valyala:
FYI, VictoriaMetrics gained support for Graphite Tags API starting from v1.47.0 - see https://victoriametrics.github.io/#graphite-tags-api-usage .

Civil:
AFAIR current version of carbonapi uses optimal VictoriaMetrics handlers to fetch data, so this one can be closed.

PR #542, opened 2020-11-13
Add special protocol for victoriametrics

Civil:
It's currently based on prometheus, but utilize /metrics/expand
that is present in VictoriaMetrics, but is not standard for
prometheus backends.

That should have significant performance improvements for all
/metrics/find queries.

Related to #521


Felixoid (approved):
I don't have a way to check it, but from the code, it looks good to me

Issue #166, opened 2016-10-06
timeShift() needs resetEnd=False


dgryski:
Consider:  `target=metric&target=timeshift(metric, '24h')&from=midnight&until=tomorrow`

that should display 1d of data from midnight to midnight, with metric graphed as much as we have, overlaid on all of yesterday's data.

dgryski:
Turns out this works.  I need to hunt down the (internal) query that was failing and resimplify it.

Civil:
Can you provide more information about that? So I'll be able to fix that.

dgryski:
Docs for timeshift http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.timeShift

IIRC we don't support `resetEnd=false`

nnuss:
> If resetEnd is False, will instead draw full range including future time.

My read of the doc is the opposite. That we always behave like `resetEnd=false`.
I'm also experiencing graphite (0.9.12) draws the same graph for **both** `false` and `true` (with undefined implying the default of `resetEnd=true`).

dgryski:
That's entirely possible :D  I just remember seeing a graph that was using `resetEnd` and noticing that our code did the WrongThing.   In either case, our code is incomplete with respect to upstream graphite-web.

Civil:
I think that was actually fixed in #543 long time ago.

PR #543, opened 2020-11-20
timeShift support resetEnd

faceair:
Ref https://github.com/go-graphite/carbonapi/issues/166

In previous versions we didn't implement resetEnd, so the default value was actually false, now I've added the resetEnd default value of true, which will cause many of our existing dashboards to break.

Can we set the default value of resetEnd to false? Although it is not compatible with graphite-web

### DIFF
--- a/expr/functions/timeShift/function.go
+++ b/expr/functions/timeShift/function.go
@@ -30,9 +30,13 @@ func New(configFile string) []interfaces.FunctionMetadata {
 
 // timeShift(seriesList, timeShift, resetEnd=True)
 func (f *timeShift) Do(ctx context.Context, e parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types.MetricData, error) {
-	// FIXME(dgryski): support resetEnd=true
 	// FIXME(civil): support alignDst
 	offs, err := e.GetIntervalArg(1, -1)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


Civil:
I think it would be better to do a config option to change the default value of the function (I can actually add it myself) and for next ~6months print a warning on start that the default will change to "true" to match graphite-web's behavior.

faceair:
Okay, then please add an option after the merge.
I'm not sure if my understanding of resetEnd is correct, I referenced [graphite-web's code](https://github.com/graphite-project/graphite-web/blob/1.1.7/webapp/graphite/render/functions.py#L4254-L4255) to implement it, so you can double check.

Also CI seems to have failed because of a problem with the build script itself, can you fix it first? https://travis-ci.org/github/go-graphite/carbonapi/jobs/744826566

Civil (approved):

Civil:
Yeah, I'll add it and will fix the test.

But the test fail is actually legit. Because of 'resetEnd' option and it's default value (true) test that's described here https://github.com/go-graphite/carbonapi/blob/main/cmd/mockbackend/testcases/pr529/pr529.yaml started to fail because it expected resetEnd=false behavior.

Civil:
For this release default value will be `false` for compatibility reasons, but it will change to `true` in next major release.

Also function will produce a warning, pointing to documentation where the config file is briefly described.

faceair:
Great! 👍
I'll also pay attention to the e2e test cases next time.

PR #544, opened 2020-12-06
victoriametrics: probe victoriametrics backend for version information

Civil:
Some APIs, like /metrics/expand are supported since specific version of VM.

Add a prober and relevant configuration options that would choose implementation
depending on VictoriaMetrics. That would allow to implement optimized query
functions and not to rely on users to have latest version of VM installed.

Related to #521


Felixoid (approved):
LGTM

PR #560, opened 2021-02-09
Return 400 on bad requests

Felixoid:
On expression errors, BadRequest status code will be returned


cmd/carbonapi/http/render_handler.go
```diff
@@ -309,7 +314,7 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
 		if returnCode == 404 {
 			returnCode = config.Config.NotFoundStatusCode
 		}
-		if returnCode >= 500 {
+		if returnCode >= 500 || returnCode == 400 {
```
Civil, line 317: I actually wonder if it won't be better to have a separate log message for error code 400

Civil (approved):

PR #561, opened 2021-02-10
Implement github action for simple tests

Felixoid:
Travis looks worse and worse each week. Here are the simplest tests

Most probably, `travis-docker-build.sh` should be adjusted for the package building


Felixoid:
Here you can see, tests work https://github.com/Felixoid/carbonapi/actions/runs/553003491

PR #569, opened 2021-02-20
Bump go version requirement to 1.14 and add test build on 1.16

Civil:
Even though carbonapi technically builds with 1.13, there is no need
to test such an old version of golang. Remove corresponding test for
github actions.

Also remove all go versions except latest from travis autobuilds as
since we've mostly migrated to github actions it just only do extra
work with no benefit

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,9 +14,9 @@ jobs:
     strategy:
       matrix:
         go:
-          - ^1.13
           - ^1.14
           - ^1.15
+          - ^1.16
           - ^1
     steps:
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,6 @@ git:
 
 matrix:
   include:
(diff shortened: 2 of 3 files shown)


Felixoid (approved):

.github/workflows/tests.yml
```diff
@@ -14,9 +14,9 @@ jobs:
     strategy:
       matrix:
         go:
-          - ^1.13
```
Felixoid, line 17: As mentioned, we may leave it until the compatibility would be dropped by some new language feature

Civil: To be honest I don't think we need to support even 1.15 (or well, maybe after 1.16.1 release it should be ok to drop 1.15) as usually people migrate to latest golang rather quickly, but I would leave 1.14 as "until something breaks" and would try to maintain compatibility with 1.15 while it's still officially supported.

Issue #570, opened 2021-02-21
[BUG]  Seems method (bg *BroadcastGroup) splitRequest  have a little typo 


tantra35:
**Describe the bug**
When we decided to switch to use `maxBatchSize: 100` from  `maxBatchSize: 0` we have see abnormal request count from carbonapi to our go-carbonservers, this count incresed up to 100 times. Its abnormal due we doesn't use so huge wildcard request, as one might think

So we inspected sources a little bit and make a little change in `splitRequest` method(by our oppinion it have a bug - typo?)

```patch
 zipper/broadcast/broadcast_group.go | 27 +++++++++++++++++----------
 1 file changed, 17 insertions(+), 10 deletions(-)

diff --git a/zipper/broadcast/broadcast_group.go b/zipper/broadcast/broadcast_group.go
index 775b9377..cde3ab26 100644
--- a/zipper/broadcast/broadcast_group.go
+++ b/zipper/broadcast/broadcast_group.go
@@ -152,9 +152,10 @@ func (bg *BroadcastGroup) doMultiFetch(ctx context.Context, logger *zap.Logger,
 		return
 	}
 
+	logger = logger.With(zap.String("backend_name", backend.Name()), zap.Int("splited requests", len(requests)))
+
 	for _, req := range requests {
 		go func(req *protov3.MultiFetchRequest) {
-			logger = logger.With(zap.String("backend_name", backend.Name()))
 			logger.Debug("waiting for slot",
 				zap.Int("max_connections", bg.limiter.Capacity()),
 			)
@@ -204,7 +205,7 @@ func (bg *BroadcastGroup) doSingleFetch(ctx context.Context, logger *zap.Logger,
 		return
 	}
 
-	logger = logger.With(zap.String("backend_name", backend.Name()))
+	logger = logger.With(zap.String("backend_name", backend.Name()), zap.Int("splited requests", len(requests)))
 	logger.Debug("waiting for slot",
 		zap.Int("max_connections", bg.limiter.Capacity()),
 	)
@@ -241,9 +242,15 @@ func (bg *BroadcastGroup) splitRequest(ctx context.Context, request *protov3.Mul
 	}
 
 	var requests []*protov3.MultiFetchRequest
+	newRequest := &protov3.MultiFetchRequest{}
+
 	var err merry.Error
 	for _, metric := range request.Metrics {
-		newRequest := &protov3.MultiFetchRequest{}
[...]

PR #571, opened 2021-02-21
Seems method (bg *BroadcastGroup) splitRequest have a little typo

tantra35:
When we decided to switch to use maxBatchSize: 100 from maxBatchSize: 0 we have see abnormal request count from carbonapi to our go-carbonservers, this count incresed up to 100 times. Its abnormal due we doesn't use so huge wildcard request, as one might think

After that we see dramaticaly reduce recouest count to our go-carbon servers, and many widlcard request begin work as expected

This patch make 2 major things:

1.    Make splitRequest reinterable on the same metric set. So if we pass already processed metrics(after expand wildcard throw find) we have got the same "batches", as after find

2.    Remove typo, by doing transformation Name: metric.PathExpression, => Name: metric.Name, on
    https://github.com/go-graphite/carbonapi/blob/main/zipper/broadcast/broadcast_group.go#L263

But this place of code https://github.com/go-graphite/carbonapi/blob/main/zipper/broadcast/broadcast_group.go#L250,  where we doubt whether fix needed

This PR created as replacement for https://github.com/go-graphite/carbonapi/issues/570

### DIFF
--- a/zipper/broadcast/broadcast_group.go
+++ b/zipper/broadcast/broadcast_group.go
@@ -241,9 +241,15 @@ func (bg *BroadcastGroup) splitRequest(ctx context.Context, request *protov3.Mul
 	}
 
 	var requests []*protov3.MultiFetchRequest
+	newRequest := &protov3.MultiFetchRequest{}
+
 	var err merry.Error
 	for _, metric := range request.Metrics {
-		newRequest := &protov3.MultiFetchRequest{}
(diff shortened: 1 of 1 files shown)


Civil (approved):

Civil:
> But this place of code https://github.com/go-graphite/carbonapi/blob/main/zipper/broadcast/broadcast_group.go#L250, where we doubt whether fix needed

I think it's not required, but it makes code easier to follow so I think that's a good thing to have anyway.

PR #574, opened 2021-02-28
Custom dns multi accept refactor

Civil:


cmd/carbonapi/main.go
```diff
@@ -66,12 +74,23 @@ func main() {
...
-				err = gracehttp.Serve(&http.Server{
-					Addr:    config.Config.Expvar.Listen,
+			go func(listen config.Listener) {
+				l := &net.ListenConfig{Control: helper.ReusePort}
+				s := &http.Server{
+					Addr:    listen.Address,
 					Handler: handler,
-				})
+				}
+				listener, err := l.Listen(context.Background(), "tcp", listen.Address)
```
Felixoid, line 87: Can we resolve an address? E.g., localhost is always resolved to `127.0.0.1` + `::1`. Then function would create `s` per resolved adress

Felixoid (approved):
LGTM

PR #578, opened 2021-03-07
Back-port functions from avito's carbonapi fork

Civil:
Port a commit that Avito did in their fork of carbonapi

This commit adds an optional parameter `allowFormatStr`
to `alias` function, if set to `True` it will replace all
all `${expr}` passed in alias string with actual original
expression.


expr/functions/sortByName/function.go
```diff
@@ -35,19 +35,30 @@ func (f *sortByName) Do(ctx context.Context, e parser.Expr, from, until int64, v
 		return nil, err
 	}
 
-	natSort, err := e.GetBoolNamedOrPosArgDefault("natural", 1, false)
+	isNatSort, err := e.GetBoolNamedOrPosArgDefault("natural", 1, false)
+	if err != nil {
+		return nil, err
+	}
+
+	isReverseSort, err := e.GetBoolNamedOrPosArgDefault("reverse", 2, false)
```
Felixoid, line 43: It makes sense to update a description

expr/functions/timeShiftByMetric/function.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func New(configFile string) []interfaces.FunctionMetadata {
+	return []interfaces.FunctionMetadata{interfaces.FunctionMetadata{
+		F:    &timeShiftByMetric{},
+		Name: "timeShiftByMetric",
+	}}
+}
+
+func (f *timeShiftByMetric) Description() map[string]types.FunctionDescription {
```
Felixoid, line 35: Just note, usually, the Description is at the end of a file. I don't insist, it may stay here to be consistent with Avito fork.

expr/functions/scale/function.go
```diff
@@ -21,7 +21,7 @@ func GetOrder() interfaces.Order {
 func New(configFile string) []interfaces.FunctionMetadata {
 	res := make([]interfaces.FunctionMetadata, 0)
 	f := &scale{}
-	functions := []string{"scale"}
+	functions := []string{"scale", "scaleAfterTimestamp"}
```
Felixoid, line 24: It will return the same Description for both names, richt?

expr/functions/sortByName/function.go
Civil: Parameter was already in description, but wasn't actually used.

expr/functions/timeShiftByMetric/function.go
Civil: Fixed

expr/functions/scale/function.go
Civil: AFAIR I need to provide description for both metrics... Added one.

COMPATIBILITY.md
```diff
@@ -105,28 +105,31 @@ _When `format=png`_ (default if not specified)
...
 
-## Graphite-web 1.1 compatibility
+## Graphite-web 1.1.7 compatibility
 ### Unsupported functions
 | Function                                                                  |
 | :------------------------------------------------------------------------ |
 | aggregateWithWildcards |
 | aliasQuery |
 | averageOutsidePercentile |
+| countSeries |
```
Felixoid, line 115: count-, diff- and multiplySeries are supported functions

Felixoid:
It may be merged after fixed COMPATIBILITY.md

PR #582, opened 2021-03-10
Honor isleaf atribute

tantra35:
this patch allow proper de duplication of metric names when some metric also have a children for example, whisper files layout:

```
put.wsp
put/
   somechild.wsp
```

without this carbonapi  will produce wrong results and some metrics for wildcard may put twice or more


Civil (approved):

tantra35:
@Civil Thanks for approve, sorry but may a ask you, why you merge so quick? without discussion

Civil:
Well, previous behavior looks like a bug to me, I had a look at a proposed fix and it looks OK. So not sure what to discuss in that case.

For most of the PRs here it's usually either me or @deniszh or @Felixoid do the review, one reviewer is enough as well to merge it.

PR #585, opened 2021-03-17
Scale input series to same value if they are passed to aggregate func…

### LINKED ISSUE #584: [BUG] panic if backend respects maxDataPoints and `time` function specified, carbonapi will panic
**Describe the bug**
Subj.

**CarbonAPI Version**
0.14.2 and later (including 0.15.x and current master).

**Simplified query (if applicable)**
```
diffSeries(time("t"), metric.name)
```

Civil:
…tion

 * Because test requires to support `from=0` fixes data-time parsing to closer match graphite's:
   https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/attime.py#L126
 * Add `ScaleToCommonStep` to aggregate helper function

Fixes #584

### DIFF
--- a/cmd/mockbackend/render.go
+++ b/cmd/mockbackend/render.go
@@ -112,13 +112,13 @@ func (cfg *listener) renderHandler(wr http.ResponseWriter, req *http.Request) {
 			time.Sleep(delay)
 		}
 		for _, m := range response.Data {
-			startTime := m.StartTime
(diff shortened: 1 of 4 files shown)


Felixoid (approved):
Looks good to me, but please, review my changes too

Civil:
LGTM then

Issue #593, opened 2021-04-02
[FR] Error handle (for check on balancers)


msaf1980:
Carbonapi return 500 code on non-recoverable error (may be all backends down). It's difficult to retry on balancers and hard to monitoring (internal or communication error).

May be better return aggregated error (in some cases 503/504) ?

PR #594, opened 2021-04-02
error handle (for check on balancers)

msaf1980:
PR for https://github.com/go-graphite/carbonapi/issues/593

Merge errors and try to choose aggregated error code.


cmd/carbonapi/http/find_handlers.go
```diff
@@ -192,7 +192,7 @@ func findHandler(w http.ResponseWriter, r *http.Request) {
 	query := r.Form["query"]
 	srcIP, srcPort := splitRemoteAddr(r.RemoteAddr)
 
-	accessLogger := zapwriter.Logger("access")
+	accessLogger := zapwriter.Logger("find")
```
Civil, line 195: Not sure if that's required.

Purpose of accessLogger is to produce exactly one message per request (as you'd expect from access logs in http servers) with all possible request stats. And it's named "access" on purpose as usually you either want all access logs or none of them. However I'm open to discuss that if you think that it would be worth to silence only access logs for specific handlers. But even in that case logs are sort of cheap (in terms of cost to call .Info in zap) so I'd suggest to add a warning on startup that behaviour will be changed soon (and only if some one have access logger redefined in config), update the doc, create a separate logger per handler and log to both access and that separate logger.

zipper/helper/requests.go
```diff
@@ -5,17 +5,114 @@ import (
...
+	errMsgs := make([]string, 0)
+	for _, err := range errors {
+		var code int
+		c := merry.RootCause(err)
+		if c == nil {
+			c = err
+		}
+		code = merry.HTTPCode(c)
+
+		if merry.HTTPCode(c) == 404 || merry.Is(c, parser.ErrSeriesDoesNotExist) {
```
Civil, line 47: even though current code doesn't have consistency across codes, but I'd suggest to try to replace all hardcoded codes as constants from net/http, e.x. 404 -> http.StatusNotFound

zipper/helper/requests.go
```diff
@@ -5,17 +5,114 @@ import (
...
+		errMsgs = append(errMsgs, c.Error())
+		if code == 400 {
+			// The 400 is returned on wrong requests, e.g. non-existent functions
+			returnCode = code
+		} else if returnCode == 404 {
+			// First error
+			returnCode = code
+		} else if code == 503 && (returnCode == 504 || returnCode == 502) {
+			returnCode = 503
+		} else if (code < 502 || code > 504) && (returnCode >= 502 || returnCode <= 504) {
```
Civil, line 59: I'm not really following the logic here and what makes 503 so special for example (it looks like this `else if` can be merged with previous one that was for 403 and as well with the one that's for 400).

I'm also not so sure if returnCode will ever be changed to something other than 400 or 404. It looks like that all changes that could've potentially set it to 504 or 503 are guarded with `(returnCode == 504 || returnCode == 502)` or with `(reutrnCode >= 502 || returnCode <= 504)`.

zipper/helper/requests.go
```diff
@@ -5,17 +5,114 @@ import (
...
+		}
+	}
+
+	return returnCode, errMsgs
+}
+
+func MergeHttpErrorMap(errors map[string]merry.Error) (int, []string) {
+	returnCode := http.StatusNotFound
+	errMsgs := make([]string, 0)
+	for _, err := range errors {
```
Civil, line 70: It looks like the for-loop here and in `MergeHttpErrors` is mostly the same, so maybe there is a way how to make one function to just convert inputs and then call the other one?

Civil:
I understand that this PR is a draft, but I left a few comments about the current state

zipper/helper/requests.go
msaf1980: Yes, logic of error handling is a main question.
May be better - priority of 400, 403, 500 and other non-retryble errors instead of communication errors (502, 503, 504).
Becouse in this case query might be wrong, access denied (possible limits in graphite-clickhouse). We don't need retry this queries.

Civil (changes requested):
And I basically cannot properly review the PR because of a lot of commits that were merged from master to the branch. Sorry.

msaf1980:
> And I basically cannot properly review the PR because of a lot of commits that were merged from master to the branch. Sorry.



> I understand that this PR is a draft, but I left a few comments about the current state



> And I basically cannot properly review the PR because of a lot of commits that were merged from master to the branch. Sorry.



> Could you please rebase this on top of current master? I really worried that github thinks there are 33 commits to be merged (and most of them were some of the recent commits to the master).

Done.

zipper/protocols/v2/protobuf_group.go
```diff
@@ -360,10 +361,10 @@ func (c *ClientProtoV2Group) doTagQuery(ctx context.Context, isTagName bool, que
 	logger := c.logger
 	var rewrite *url.URL
 	if isTagName {
-		logger = logger.With(zap.String("type", "tagName"))
+		logger = logger.With(zap.String("type", "tagName"), zap.String("context_uuid", utilctx.GetUUID(ctx)))
```
Civil, line 364: I would name that field not a `context_uuid` but rather `carbonapi_uuid` to be consistent with all other places in carbonapi.

Civil (changes requested):

Civil:
I will try to play with the change tomorrow and maybe I'll have more comments about the error handling logic.

cmd/carbonapi/http/errors.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func HttpCode(err error, found bool) int {
+	if err == nil {
+		if !found {
+			return http.StatusNotFound
+		}
+		return http.StatusOK
+	}
+	if _, ok := err.(*net.OpError); ok {
```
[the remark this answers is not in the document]
msaf1980: It's for low level error handling (from standard golang http client). No merry at this time. Yes, It can be wrapped. But it's encapsulated with some parent errors, like ErrFailedToFetch or ErrMaxTriesExceeded. So I can't easy extract merry error, but can extract original (root, not merry) error.

zipper/helper/requests.go
```diff
@@ -120,7 +219,15 @@ func (c *HttpQuery) doRequest(ctx context.Context, logger *zap.Logger, server, u
 		logger.Debug("error fetching result",
 			zap.Error(err),
 		)
-		return nil, merry.Here(err).WithValue("server", server)
+
+		if urlErr, ok := err.(*url.Error); ok {
```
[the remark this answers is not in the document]
msaf1980: Errors checks moved to function. Errors can be wrapped with needed marry errors, but no simple way to extract child (first) merry error. Loop before root (no-merry) error. Not elegant and can produce errors in future (if we not set correct merry wrapper error at first). Current scheme handling all low-level errors in one place, and it's more easy to undestand, than work with error levels with merry (it's more difficult and probably baggy).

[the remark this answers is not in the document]
msaf1980: I replace error checking with more simple model with status codes.

Civil:
@Felixoid can you also have a look at this PR as I might be missing something.

Felixoid (commented):
Everything is fine, just one question

zipper/helper/requests.go
```diff
@@ -5,17 +5,109 @@ import (
...
+		}
+		if msg := merry.Message(c); len(msg) > 0 {
+			errMsgs = append(errMsgs, strings.TrimRight(msg, "\n"))
+		} else {
+			errMsgs = append(errMsgs, c.Error())
+		}
+
+		if code == http.StatusGatewayTimeout || code == http.StatusBadGateway {
+			// simplify code, one error type for communications errors, all we can retry
+			code = http.StatusServiceUnavailable
```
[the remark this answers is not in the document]
msaf1980: http.StatusServiceUnavailable not set from other code directly. It's must be returned by upstream (gch for example) or masked from http.StatusGatewayTimeout or http.StatusBadGateway for simplify error handling (it's can be retry on balancers).

Directly set only some errors like timeout, handled on https://github.com/msaf1980/carbonapi/blob/error_check/zipper/types/response.go#L59
https://github.com/msaf1980/carbonapi/blob/error_check/zipper/types/errors.go#L20.

It's masked to http.StatusServiceUnavailable 
https://github.com/msaf1980/carbonapi/blob/error_check/zipper/helper/requests.go#L60

Felixoid (approved):

Felixoid:
@Civil There is still `changes requested` status from you

Civil (approved):

Issue #606, opened 2021-05-13
[BUG]  Enable cairo/png support by default in the Docker image


travis-bear:
**Describe the bug**
Currently the docker image for carbonapi is built with cairo/png support explicitly disabled.

**CarbonAPI Version**
Docker image `gographite/carbonapi:v0.15.1`

**CarbonAPI Configuration:**
```
listen: "0.0.0.0:8081"

# Specify URL Prefix for all handlers
prefix: ""

# Specify if metrics are exported over HTTP and if they are available on the same address or not
# pprofEnabled controls if extra HTTP Handlers to profile and debug application will be available
expvar:
  enabled: true
  pprofEnabled: false
  listen: ""

# Controls headers that would be passed to the backend
headersToPass:
  - "X-Dashboard-Id"
  - "X-Grafana-Org-Id"
  - "X-Panel-Id"
headersToLog:
  - "X-Dashboard-Id"
  - "X-Grafana-Org-Id"
  - "X-Panel-Id"
# Specify custom function aliases.
# This is example for alias "perMinute(metrics)" that will behave as "perSecond(metric)|scale(60)"
define:
  - name: "perMinute"
    template: "perSecond({{.argString}})|scale(60)"
# Control what status code will be returned where /render or find query do not return any metric. Default is 200
notFoundStatusCode: 200

# Max concurrent requests to CarbonZipper
concurency: 1000
cache:
  # Type of caching. Valid: "mem", "memcache", "null"
  type: "mem"
  # Cache limit in megabytes
  size_mb: 0
  # Default cache timeout value. Identical to DEFAULT_CACHE_DURATION in graphite-web.
  defaultTimeoutSec: 60
  # Only used by memcache type of cache. List of memcache servers.
#   memcachedServers:
#       - "127.0.0.1:1234"
#       - "127.0.0.2:1235"
# Amount of CPUs to use. 0 - unlimited
cpus: 0

maxBatchSize: 100
graphite:
  host: "relay-graph.<...>:2003"
  interval: "60s"
  prefix: "carbon.api"
idleConnections: 10
pidFile: ""

upstreams:
  tldCacheDisabled: false
  buckets: 10
  slowLogThreshold: "3s"
  timeouts:
    find: "4s"
    render: "15s"
    connect: "600ms"

  concurrencyLimitPerServer: 0
  keepAliveInterval: "30s"
  maxIdleConnsPerHost: 100
  doMultipleRequestsIfSplit: false

  backendsv2:
[...]

PR #609, opened 2021-05-16
Including cairo in docker build (fixing #606) 

deniszh:
Also adding RHEL logrotate file


Felixoid (approved):
Looks perfect, thank you!

Issue #599, opened 2021-04-22
[FR] Drop user privileges


tleuxner:
**Is your feature request related to a problem? Please describe.**
Currently **carbonapi** starts its service as root under _systemd_.

**Describe the solution you'd like**
Carbonapi should drop elevated privileges and run under an unprivileged user.

**Additional context**
My current workaround is to use an override for systemd. However I reckon a general setting for all init platforms would be a better approach.

**/etc/systemd/system/carbonapi.service.d/override.conf**
```
[Service]
User=carbon
Group=carbon
```

Civil:
Problem here is that this change seems to be a breaking one (or potentially breaking one) as people who use packages already will need to have some migration path (as an obvious problem - logs will be owned by root).

I'm also not really sure how to properly handle all of that with `nfpm` (I use it instead of proper package files to build packages).

Otherwise - yeah, that's a good point of course and eventually should be changed to a less privileged user.

Civil:
That will be fixed once the https://github.com/go-graphite/helper-scripts/pull/4 will be merged (before next release).

PR #610, opened 2021-05-17
WIP: Carbon user implementation

deniszh:
Fixing #599
Require changes in helper-scripts.

### DIFF
--- a/contrib/carbonapi/fpm/carbon-user-systemd-reload.sh
+++ b/contrib/carbonapi/fpm/carbon-user-systemd-reload.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+# based on https://github.com/leoleovich/grafsy/blob/master/packaging/postinst
+
+# Abort if any command returns an error value
+set -e
+
+USER=carbon
+GROUP=carbon
+CONF=/etc/carbonapi/carbonapi.yaml
+PIDFILE=/var/run/carbonapi/carbonapi.pid
+LOGFILE=/var/log/carbonapi/carbonapi.log
+
+# Following user part should be tested on both RPM and DEB systems
+if ! getent group "${GROUP}" > /dev/null 2>&1 ; then
+  groupadd --system "${GROUP}"
+fi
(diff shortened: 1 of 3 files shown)


contrib/carbonapi/fpm/carbon-user-systemd-reload.sh
```diff
@@ -0,0 +1,45 @@
...
+GID=$(getent group "${GROUP}" | cut -d: -f 3)
+if ! id "${USER}" > /dev/null 2>&1 ; then
+  adduser --system --home /dev/null --no-create-home \
+    --gid "${GID}" --shell /bin/false \
+    "${USER}"
+fi
+
+# fix PID permissions
+if [ -f "${PIDFILE}" ]; then
+    chown "${USER}":"${GROUP}" "${PIDFILE}"
```
deniszh, line 27: Not sure if that's good idea. Theoretically, we should check ownership and permission of pid and logfile and change it only if not com patible with current version - but that make script quite complex....

contrib/carbonapi/fpm/carbon-user-systemd-reload.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [ ! -e "${CONF}" ]; then
+  echo "For use this software you have to create ${CONF} file. You could use /usr/share/carbonapi/carbonapi.example.yaml as default"
+else
+  # On debian jessie (systemd 215) it fails if symlink already exists
+  systemctl is-enabled carbonapi || systemctl enable carbonapi
+  # Check if systemd is up and running, e.g. not in chroot
+  if systemctl 1>/dev/null 2>&1; then
+    systemctl daemon-reload
+    systemctl restart carbonapi.service
```
Civil, line 43: I personally don't like the idea of restarting service in postinst scripts without a way how to disable that.

deniszh: Good point. I removed enabling also, let's keep only reload there.

Civil (approved):

PR #613, opened 2021-05-25
allow custom fpm-like tool and equal version for all build scripts

msaf1980:
Changes in this PR:
1) Allow to choose custom fpm-like cli tool (no Ruby dependency for build)
2) Equal version for carbonapi/carbonzipper packages (RPM/DEB)


Civil:
Those are the old scripts that I've used to create packages, current version is in another repository and uses nfpm (fpm-alternative written in Go): https://github.com/go-graphite/helper-scripts

But anyway that's overall a good idea to keep them in a good shape.

Civil (approved):

Issue #616, opened 2021-05-28
[BUG] Cached values can be overwritten in evaluator.FetchAndEvalExp


msaf1980:
For example

target: alias(averageSeries(test.*.events), 'avg')
target: aliasByNode(test.*.events, 4)

On first target evalution value name was overwritten (in averageSeries, see https://github.com/go-graphite/carbonapi/blob/main/expr/helper/helper.go#L159, but in some other functions name will be overwritten too)
So,second target evaluation will work with incorect values from cache (passed in https://github.com/go-graphite/carbonapi/blob/main/expr/expr.go#L89)

Deep copy when pass values from cache solved this

https://github.com/msaf1980/carbonapi/blob/fix_cachemod/expr/expr.go#L89

But it's might be too expansive for large dataset.
But work with individual functions not a simple and probably wrong way, too easy forgot a copy.

PR #618, opened 2021-05-28
Fix eval copy on write

msaf1980:
Fix for https://github.com/go-graphite/carbonapi/issues/616


Felixoid (approved):
Thank you for fixing it :+1:

PR #619, opened 2021-06-01
Fix extract node

msaf1980:
Strip function/args from name tag in helper.AggKey.
Affected function:
```
aliasByNode
weightedAverage
```

`groupByNode` not affected, because not use helper.AggKey.


msaf1980:
Problem was more complex and changes in PR not work as expected.

Issue #621, opened 2021-06-14
[BUG] victoriametrics version detection fails for vm cluster


ikod:
**Describe the bug**

carbonapi detects VM version quering 'metrics' url formed from the backend servers base url just appending '/metrics'. This works both for single node and cluster version, as VM supports url http://vmnode/metrics for both cluster and single-node. See https://github.com/go-graphite/carbonapi/blob/main/zipper/protocols/victoriametrics/feature_set.go#L149

For `query_range` cluster VM require `AccountID` in request url: http://vmnode/select/0/prometheus/query_range... so 
base url in `backends` must include this `/select/0/prometheus` part. But clustered VM do not support `metrics` request like http://vmnode/select/0/prometheus/metrics - which carbonapi will use to detect VM version.

_So the problem is: when working with clustered VM, carbonapi can either detect VM version, or make timeseries queries, not both._

**CarbonAPI Version**

v0.15.4

**Logs**
```
2021-06-14T07:20:38.775Z        DEBUG   zipper  got slot for server     {"type": "victoriametrics", "protocol": "victoriametrics", "name": "victoriametrics", "function": "updateFeatureSet", "function": "HttpQuery.doRequest", "server": "http://vmselect-cluster.internal/select/0/prometheus", "name": "victoriametrics", "uri": "http://vmselect-cluster.internal/select/0/prometheus/metrics", "name": "http://vmselect-cluster.internal/select/0/prometheus"}

[...]

PR #622, opened 2021-06-19
add support for victoriametrics cluster version

ikod:
fix for https://github.com/go-graphite/carbonapi/issues/621

### DIFF
--- a/zipper/protocols/victoriametrics/fetch.go
+++ b/zipper/protocols/victoriametrics/fetch.go
@@ -26,8 +26,8 @@ func (c *VictoriaMetricsGroup) Fetch(ctx context.Context, request *protov3.Multi
 	logger := c.logger.With(zap.String("type", "fetch"), zap.String("request", request.String()))
 	stats := &types.Stats{}
 	var serverUrl string
-	if c.vmClusterTenantId >= 0 {
-		serverUrl = fmt.Sprintf("http://127.0.0.1/select/%d/prometheus/api/v1/query_range", c.vmClusterTenantId)
+	if len(c.vmClusterTenantID) > 0 {
+		serverUrl = fmt.Sprintf("http://127.0.0.1/select/%s/prometheus/api/v1/query_range", c.vmClusterTenantID)
 	} else {
 		serverUrl = "http://127.0.0.1/api/v1/query_range"
 	}
--- a/zipper/protocols/victoriametrics/fetch.go
+++ b/zipper/protocols/victoriametrics/fetch.go
@@ -25,7 +25,13 @@ func (c *VictoriaMetricsGroup) Fetch(ctx context.Context, request *protov3.Multi
 	}
 	logger := c.logger.With(zap.String("type", "fetch"), zap.String("request", request.String()))
 	stats := &types.Stats{}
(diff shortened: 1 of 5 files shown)


ikod:
Maybe some explanation related to VMCluster vs VMSingle in the docs required, but I can't choose right place.

deniszh:
Hi @ikod ,

You can add documentation for `vmClusterTenantId` option near `backendOptions` section of https://github.com/go-graphite/carbonapi/blob/main/doc/configuration.md

zipper/protocols/victoriametrics/fetch.go
```diff
@@ -25,7 +25,13 @@ func (c *VictoriaMetricsGroup) Fetch(ctx context.Context, request *protov3.Multi
 	}
 	logger := c.logger.With(zap.String("type", "fetch"), zap.String("request", request.String()))
 	stats := &types.Stats{}
-	rewrite, _ := url.Parse("http://127.0.0.1/api/v1/query_range")
+	var serverUrl string
+	if c.vmClusterTenantId >= 0 {
```
deniszh, line 29: I think it should be > 0? Or you should put "-1" as default in example config, which should work out of the box for non-clustered version IMO

ikod: Hi, @deniszh 

Tenant ID 0 is legal id, see https://docs.victoriametrics.com/Cluster-VictoriaMetrics.html#multitenancy :

* Each accountID and projectID is identified by an arbitrary 32-bit integer in the range [0 .. 2^32). If projectID is missing, then it is automatically assigned to 0

BTW, it actually called accountID, and also can have projectId. Maybe it is worth to rename it and add support for projectId?

Re "-1" as default for non-clustered version: user can omit this vmClusterTenantId option completely, and it will work as before, so upgrade would be safe with old configs, so I don't think it is worth to mention this option for single node vesrion/

deniszh: Sorry, I have not many experience with VM, so, please bear with me.
So, URL "http://127.0.0.1/select/0/prometheus/api/v1/query_range" will work even for single node VM, same as "http://127.0.0.1/api/v1/query_range" ?

deniszh (approved):

zipper/protocols/victoriametrics/fetch.go
ikod: hi @deniszh 

> So, URL "http://127.0.0.1/select/0/prometheus/api/v1/query_range" will work even for single node VM

No.

Carbonapi detect VM version using url like http://127.0.0.1/metrics. This works both in single- and cluster-mode regardless of `vmClusterTenantID` option. But for query_range we have to use accountID in API calls for **cluster mode** only.

If user **do not** use vmClusterTenantID in carbonapi `config yaml`, then carbonapi will build single node VM API url: http://127.0.0.1/api/v1/query_range

If user **use** vmClusterTenantID in config yaml then (with this PR merged) cluster API will be used http://127.0.0.1/select/accountID/prometheus/api/v1/query_range for series.

Hope this helps.

BTW I changed code a bit so tenant ID can be configured more flexible, with both accoutID ant projectID.

Civil (approved):

ikod:
Thanks!

PR #625, opened 2021-06-28
Write carbonapi uuid in response header

msaf1980:
Write carbonapi UUID in response header.


Felixoid (approved):
I'd like to see some tests for it. Most probably, a good place for them is integration tests.

Besides that, looks good, so early approval here.

Do you have time/effort to do so now or should I merge it?h

msaf1980:
Sorry, I haven't free time at now. What's testing is needed ? Check correct headers for all query types ?

Felixoid:
Yep, testing that handlers return the header with a particular format.

PR #629, opened 2021-07-09
IRONdb support for Zipper 

deniszh:
This PR is replacing PR#626.
This PR is adding support for IRONdb protocol (for reading Graphite-compatible metrics from IRONdb) to carbonzipper. PR also contains example config and description of irondb-specific backend options. Also it contains some amends for documentation for supported protocol and backend versions.

PS: for simplifying review all documentation changes grouped in commit 60f5489, all implementation files grouped to single commit b29aebf, other commit contains mod and vendor changes. If you have any questions - please let me know in comments.

Thanks a lot !


zipper/protocols/irondb/irondb_group.go
```diff
@@ -0,0 +1,682 @@
...
+							PathExpression:    pathExpr,
+							ConsolidationFunc: "Average",
+							StartTime:         realStart,
+							StopTime:          realStop,
+							StepTime:          step,
+							Values:            values,
+							XFilesFactor:      0.0,
+						})
+					}
+				}
```
Felixoid, line 405: This is a personal preference, but I'd rather use `continue` here and get rid of `else` for reducing the nesting level.

zipper/protocols/irondb/irondb_group.go
```diff
@@ -0,0 +1,682 @@
...
+				zap.String("target", target),
+				zap.Int64("start", start),
+				zap.Int64("stop", stop),
+				zap.Int64("count", count),
+				zap.Int64("period", step),
+			)
+
+			var query string
+			if strings.HasPrefix(target, "seriesByTag") {
+				query = target[len("seriesByTag(") : len(target)-1]
```
Felixoid, line 318: Maybe, better use a const for `len("seriesByTag(")`. Otherwise, it will allocate string and int for each target

Felixoid (approved):
Few points to consider. The other looks good.

deniszh:
Thanks for review, @Felixoid ! Totally valid concerns btw, but let me address that in next PR, to not mix up things again. :)

Felixoid:
щл

PR #636, opened 2021-07-21
Do not apply moving functions when step is greater than window size

### LINKED ISSUE #522: [BUG] Moving median function with an interval less than the interval of the data causes no data to appear
When you you apply the moving median function to a series of data with a larger interval than the moving median is over it gives back no data.

For example: 

If the data has points every hour, and you apply `movingMedian(series, '5min')` to it. It causes no data to be returned. Instead it should just return the original data with no medians.

This is causing issues for us when using Grafana, the 1 minutely data for small time range queries needs to have moving median applied, but if the user changes to a larger time range it will use hourly data, causing the moving median function block all of the data from the user.

jonasbleyl:
Fixes #522 

This fix is particularly useful when using Prometheus or VictoriaMetrics as the backend.

If you have a function like `movingMedian(metric1, "2min")` and then change the Grafana dashboard time range to something large (e.g. 7 days) CarbonAPI will query the backend with a step that is greater than 2min which causes the function to return an error.

This change instead makes the moving functions return the data unmodified so that queries will work regardless of step changes.

### DIFF
--- a/expr/functions/moving/function.go
+++ b/expr/functions/moving/function.go
@@ -6,6 +6,10 @@ import (
 	"math"
 	"strconv"
 
+	"github.com/lomik/zapwriter"
+	"github.com/spf13/viper"
+	"go.uber.org/zap"
+
 	"github.com/go-graphite/carbonapi/expr/helper"
 	"github.com/go-graphite/carbonapi/expr/interfaces"
 	"github.com/go-graphite/carbonapi/expr/types"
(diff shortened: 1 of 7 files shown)


Civil:
As far as I remember that is not actually a bug, becasue graphite-web returns array of NaNs in that case and we try to follow it's behavior whereever is possible.

I understand that different backends might have different approach to the same problem, but in that case I would suggest to introduce a config flag that flips a behavior (with default that mimics graphite-web).

expr/functions/moving/function.go
```diff
@@ -90,38 +90,35 @@ func (f *moving) Do(ctx context.Context, e parser.Expr, from, until int64, value
 	for _, a := range arg {
 		r := *a
 		r.Name = fmt.Sprintf("%s(%s,%s)", e.Target(), a.Name, argstr)
+
+		if windowSize == 0 {
+			result = append(result, &r)
```
Civil, line 95: I would suggest to have a config flag here. That either can be a per-function one (e.x. `return_nans_if_step_mismatch`, by default set to True and that will populate results with NaNs) or a global one with some similar name.

Otherwise that change, while making results more expected, would break compatibility with graphite-web.

Also it might be a good idea to open a bug in graphite-project/graphite-web just for them to also consider changing the behavior of the functions.

Civil (changes requested):

jonasbleyl:
Ensuring that default behaviour matches graphite-web makes sense. I've added function specific config so that this is now configurable. 

`movingMedian` previously errored instead of returning NaNs so I've fixed that as well.

I've added tests for the default behaviour but couldn't see an example of how to test functions with different functionsConfig. Do you have an example for this or is it enough to handle the default case?

jonasbleyl:
Also, DeepSource check seems to be failing due to missing keys in structs. Should I fix that? Seems like it would apply for the tests for every function though.

Civil:
Sorry for a late reply. For the tests I would ignore deepsource errors (it's actually used to to exclude test files, not sure what went wrong this time).

Civil (approved):

jonasbleyl:
Thanks @Civil! Much appreciated

PR #656, opened 2021-11-10
cache: pass config by reference for fix cacheConfig.ShortTimeoutSec

msaf1980:
Pass config by reference for fix cacheConfig.ShortTimeoutSec if cacheConfig.DefaultTimeoutSec < cacheConfig.ShortTimeoutSec


Civil (approved):

Civil:
As discussed over Telegram:
config sanitisation was performed on a copy of a cache config instead of reference and results of that were lost. That PR fixes that.

PR #660, opened 2021-12-23
Fix removeEmptySeries function

jonasbleyl:
When adding support for xFilesFactor for the removeEmptySeries function (commit 4348948) it incorrectly changed the behaviour for the default case for this function (xFilesFactor=0). It doesn't remove series even if all of their values are null. The same applies for the removeZeroSeries which doesn't delete the series even if all values are 0.

This change fixes that and now matches the graphite-web behaviour.


Civil (approved):

Civil:
Thanks!

Issue #245, opened 2017-10-02
Implement /expand method


Civil:
It's present In graphite-web 1.0+ and allows to do more flexible lookup queries.

towolf:
This new thing wants to use `/metrics/expand`

https://github.com/Icinga/icingaweb2-module-graphite

bzed:
Any plan to implement this?

Civil:
Eventually - yes, but I don't know when I'll have time to look at it. Alternatively I would be happy to review the PR that implements it.

martialblog:
Hi, 

I think this issue is solved with https://github.com/go-graphite/carbonapi/pull/748 and https://github.com/go-graphite/carbonapi/pull/865 

Thanks for the quick merge of my PR!

Grüße,
Markus

PR #663, opened 2022-01-11
partial support expand

guidao:
Implement issue: https://github.com/go-graphite/carbonapi/issues/245


Civil:
I guess as the title have "WIP" in the name it is considered to be work-in-progress (that means I won't review it in-depth).

One note, however. Current PR expect backend to have native expand handler. However I currently do not aware of any backend that implements that.

If you have your own custom backend, please consider implementing some sort of feature flags that would mark that backend implements that type of query. For example carbonapi_v3_pb have a support for Capability Message, so you should send a PR to protocols repo here adding there ExpandMsg support (also feel free to add specific data types to make it easier to use, of course if it's actually needed) and then in your PR bump a revision of protocols repo required and guard that handler with capability request (like how it's done for Victoria Metrics backend)

Civil:
Alternatively it might be worth do a find request and post-process the response, if of course find message have enough data.

guidao:
Thank you for your comments. I would like to implement this feature, but have some confusion and would like to ask for your help.
1. The current implementation is copying most of the code of the `Find` function to implement a new `Expand` function, so I am using the `MultiGlobRequest ` structure. So is reusing the `MultiGlobRequest` structure an acceptable or comparable way to do this? Or is it better to define a new similar structure?
2. `However I currently do not aware of any backend that implements that. ` Does this mean a backend such as `graphite`? But doesn't graphite support the expand method? https://graphite.readthedocs.io/en/latest/metrics_api.html#metrics-expand

Translated with www.DeepL.com/Translator (free version)

Civil:
>  structure an acceptable or comparable way to do this? 

I guess in that case it is ok to use the same structure as for find.

> By default the find request does not get enough information

In that case the most effective approach would be to implement that on a backend side and indeed use a different handler on a backend.

Generic approach here is also possible but it would be less effective, e.x. doing requests for each and every glob that you see and implement all that logic on carbonapi side.

Overall it might be easier to approach it from a backend side (I'm not sure what you are using - go-carbon or graphite-clickhouse or something else as a backend).

Once you have a working backend it will be easier to test your changes on carbonapi side.

guidao:
DeepSource execution failed, but it was not introduced by this pr. Should I fix it?

Civil:
About DeepSource - sometimes it provide false-positive (pointing to problems that were already there or ignoring exclusion of vendored dependencies). In that case you can leave it as-is.

However sometimes it detects something new - in that case it would be great if you'll be able to fix it before you mark the PR as ready.

guidao:
Got it. I will only fix the problem introduced by this pr in order to avoid too many changes.

guidao:
@Civil This pr is ready, can you help to review it?

Civil:
I will likely have some time reading it during the weekend and I'll try not to forget that I have it waiting for the review.

deniszh:
@guidao: Are you OK if you pick up your PR? If you have any updates - it would be great. Thanks!

guidao:
> @guidao: Are you OK if I pick up your PR? If you have any updates - it would be great. Thanks!

Please feel free to do so. We solved this problem using other means, so there is nothing new in this PR.

deniszh:
FYI here: IMO this approach is very flexible and gives ability to override expand in every backend but IMO it's a bit overkill, I'll create separate PR similar to https://github.com/bookingcom/carbonapi/pull/449 - which will just implement separate encoder for expand results, but results will be collected by same call as find does. IMO that should be good enough implementation.

Civil:
Thanks. Yeah, as it seems not a lot of people need it, it doesn't need to be efficient.

deniszh:
It has some usecases, like https://github.com/grafana/grafana/pull/33694 - but not widely used, no.

deniszh:
Closed in the favor of https://github.com/go-graphite/carbonapi/pull/748

guidao:
> FYI here: IMO this approach is very flexible and gives ability to override expand in every backend but IMO it's a bit overkill, I'll create separate PR similar to [bookingcom/carbonapi#449](https://github.com/bookingcom/carbonapi/pull/449) - which will just implement separate encoder for expand results, but results will be collected by same call as find does. IMO that should be good enough implementation.

It's been a bit long and my memory is a bit fuzzy, I remember I also tried to implement Expend through the Find interface. But there was some semantics that could not be overridden. Does it support expand(*.servers.*.cpu) in this way?

deniszh:
@guidao : well, in graphite-web it's implemented in exactly that way - just another wrapper around same results which find returns. Could you please elaborate why  it should have different semantics and how exactly `expand(*.servers.*.cpu)` should be supported?

deniszh:
PS: you probably mean https://github.com/grafana/grafana/pull/33694 ? I implemented having exact that use case in mind: `find?query=*.servers.*.cpu` will return `cpu`, `expand?query=*.servers.*.cpu` will return full names.

guidao:
> PS: you probably mean [grafana/grafana#33694](https://github.com/grafana/grafana/pull/33694) ? I implemented having exact that use case in mind: `find?query=*.servers.*.cpu` will return `cpu`, `expand?query=*.servers.*.cpu` will return full names.

Good, I also think the new PR is better.

PR #666, opened 2022-03-11
fix pass context to zipper

faceair:
Hi, I found that the current context is not passed correctly, and the context is lost when the function is nested.
image

Here I'll give an example of a query where the function after rewrite is not able to get maxDataPoints.
```
aliasByNode(highestAverage(applyByNode(host.dns[0-9]*.uptime.uptime.value,1,"scale(perSecond(%.interface-bond1.if_octets.tx), 0.000008)"), 5), 1, 2)
```
So I changed the signature of this method in this pr so that the context is passed correctly.

Also I have a suggestion, can we pass maxDataPoint in the Eval parameter in interfaces. Something like this.
```
Eval(ctx context.Context, e parser.Expr, from, until, maxDataPoints int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types. MetricData, error)
```


Civil (approved):

Civil:
Thanks! Merged.

PR #671, opened 2022-05-25
Fix bashism

Glandos:
According to shellcheck, `[[` can be unavailable. And indeed, on my Debian box, it is not valid with `/bin/sh`

### DIFF
--- a/contrib/carbonapi/fpm/carbon-user-systemd-reload.sh
+++ b/contrib/carbonapi/fpm/carbon-user-systemd-reload.sh
@@ -37,4 +37,6 @@ if [ ! -e "${CONF}" ]; then
 fi
 
 # reload systemd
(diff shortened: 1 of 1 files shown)


Felixoid (approved):
Looks good

Civil:
I'd rather replace shebang with `/usr/bin/env bash` to be honest, rather than get rid of bash-specific constructs in scripts.

Glandos:
Yes that's another potential move. Debian chose to use dash for better performance, and since the script was vastly POSIX shell compatible, I preferred to modify this simple test. But it's up to you :)

Felixoid:
Even with the fact that `bash` is very popular, it's still not everywhere.

I personally prefer portability over simplicity, so rather vote for POSIX compatible syntax if it doesn't harm.

PR #681, opened 2022-08-09
Fix extract tags from seriesByTag

msaf1980:
Fix extract tags from seriesByTag args (actual for aggregation functions like sumSeries, which can't correct extract tags from metrics tags correctly).
Also has some differences from current graphite-web  (which used tags from first metric).
In wildcarded/regexp args special symbols are replaced with some underscores.

### DIFF
--- a/cmd/carbonapi/carbonapi.example.clickhouse.yaml
+++ b/cmd/carbonapi/carbonapi.example.clickhouse.yaml
@@ -32,6 +32,9 @@ cpus: 0
 # Timezone, default - local
 tz: ""
 
+# By default, functions like aggregate inherit tags from first series (for compatibility with graphite-web)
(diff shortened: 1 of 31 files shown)


Civil (approved):

npazosmendez:
I think graphite-web uses the intersection of all tags rather than the tags of the first series (which makes more sense):

https://github.com/graphite-project/graphite-web/blob/ed08fa4e2f6526ae11e55f88f70d0b6d36e4052f/webapp/graphite/render/functions.py#L245-L247


I'm bringing up this here because we worked on fixing that (see https://github.com/grafana/carbonapi/pull/22/files#diff-d463a03dd2f8dc196fa2f7838f56fbc4a7723fe6dcb47cf331f6c1052fee28e7) and while merging a conflict with this change popped up. We will be sending our first PR soon!

PR #686, opened 2022-09-07
Merge Grafana's fork changes

npazosmendez:
Hello! As discussed, we are sending a PR to include the changes we've been working on [our fork ](https://github.com/grafana/carbonapi). We acknowledge this is a big and hard to review PR, merging our changes in bulk like this is not our intended workflow. This huge PR should be a one-time thing.

Since this has so many changes, we think it would be better if it was merged using the commit strategy.

## About the changes

We've been focusing on the rendering functionality of carbonapi, so our changes are mostly in the `expr` package. Roughly, our changes can be divided into 3:

- Adding support to several new functions (`aggregateWithWildcards`, `exp`, `verticalLine` and more)
- Bugfixes that came from heavily testing carbonapi with real queries and data for quite some time. Some of these were panics and some we identified as wrong results. Example: https://github.com/grafana/carbonapi/pull/53
- Adding/modifying tags/names to better match GraphiteWeb. Example: https://github.com/grafana/carbonapi/pull/22/files

## Extra notes

- We've also manually implemented the `pidfile` functionality and removed the  `github.com/facebookgo/pidfile` dependency
- There are two issues pending to be addressed that are related to conflicts we've had when merging changes from the upstream. They have to do with failing tests that we had to comment out:
    - https://github.com/grafana/carbonapi/issues/68
    - https://github.com/grafana/carbonapi/issues/65

    If you agree, we can fix these with PRs sent here directly, once this one is merged.


README.md
```diff
@@ -122,7 +122,7 @@ Golang compatibility matrix:
 
 Overall rule of thumb is that carbonapi supports last 2 major go versions. E.x. at this moment Go 1.19 and 1.18 are supported.
 
-You can verify current versions that are being tested in [CI Configuration](https://github.com/go-graphite/carbonapi/blob/main/.github/workflows/tests.yml#L14).
+You can verify current versions that are being tested in [CI Configuration](https://github.com/grafana/carbonapi/blob/main/.github/workflows/tests.yml#L14).
```
npazosmendez, line 125: Will fix

Civil:
Yeah, I understand that. I think we actually discussed that it's totally ok to do that.

I'll wait for the tests to finish and for a minor fix for the readme file, after that I'll have a closer look (won't promise that it will happen today, but I'll likely have time tomorrow morning ZRH time)

> If you agree, we can fix these with PRs sent here directly, once this one is merged.

Yeah, sure.

.github/workflows/tests.yml
```diff
@@ -9,12 +9,13 @@ jobs:
...
     name: Test code
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         go:
           - ^1.18
-          - ^1.19
+          # FIXME https://github.com/grafana/carbonapi/issues/70
+          # - ^1.19
```
msaf1980, line 18: Can be removed. Issue was fixed

Civil: I can remove that later myself, just not to block the PR

expr/functions/aliasByBase64/function.go
```diff
@@ -46,7 +46,8 @@ func (f *aliasByBase64) Do(ctx context.Context, e parser.Expr, from, until int64
 		if withoutFieldArg {
 			decoded, err := base64.StdEncoding.DecodeString(a.Name)
 			if err == nil {
-				r = a.CopyName(string(decoded))
+				r = a.CopyLink()
```
msaf1980, line 49: This changes saves original tags. But it's really needed after alias ? May be only name tag ?

expr/functions/aliasByMetric/function.go
```diff
@@ -30,7 +31,7 @@ func New(configFile string) []interfaces.FunctionMetadata {
 
 func (f *aliasByMetric) Do(ctx context.Context, e parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types.MetricData, error) {
 	return helper.ForEachSeriesDo1(ctx, e, from, until, values, func(a *types.MetricData) *types.MetricData {
-		metric := a.Tags["name"]
+		metric := metric.ExtractMetric(a.Name)
```
msaf1980, line 34: Really needed ? It's more cost then simple use tag name. which already contain all, that we need (because all functions must keep tags consistent).

expr/consolidations/consolidations.go
```diff
@@ -195,11 +198,14 @@ func SummarizeValues(f string, values []float64, XFilesFactor float32) float64 {
 		rv = Percentile(values, 50, true)
 		total = notNans(values)
 	case "multiply":
-		rv = values[0]
-		for _, av := range values[1:] {
-			if !math.IsNaN(av) {
+		rv = 1.0
+		for _, v := range values {
+			if math.IsNaN(v) {
```
msaf1980, line 203: May be add tests for this ?

expr/consolidations/consolidations.go
```diff
@@ -375,19 +381,30 @@ func AggCount(v []float64) float64 {
...
 }
 
-// AggCount counts non-NaN points
 func AggDiff(v []float64) float64 {
-	res := v[0]
-	if len(v) == 1 {
-		return res
-	}
-	for _, vv := range v[1:] {
+	safeValues := make([]float64, 0)
```
msaf1980, line 385: It's can produce high memory allocations.

expr/consolidations/consolidations.go
```diff
@@ -375,19 +381,30 @@ func AggCount(v []float64) float64 {
...
 
-	return res
+	if len(safeValues) > 0 {
+		res := safeValues[0]
+		if len(safeValues) == 1 {
+			return res
+		}
+
+		for _, vv := range safeValues[1:] {
+			if !math.IsNaN(vv) {
```
msaf1980, line 399: Check really needed ?

msaf1980: May be simplified with one-pass scan and boolean flags is all is NaN ? It's more simple and has better performance (no copy slice elements).

expr/functions/aliasByNode/function_test.go
```diff
@@ -124,14 +131,15 @@ func TestAliasByNode(t *testing.T) {
...
 		},
-		{
-			Target: "aliasByNode(metric1.fo*.bar.baz,1,3)",
-			M: map[parser.MetricRequest][]*types.MetricData{
-				{Metric: "metric1.fo*.bar.baz", From: 0, Until: 1}: {types.MakeMetricData("metric1.foo==.bar.baz", []float64{1, 2, 3, 4, 5}, 1, now32)},
-			},
-			Want: []*types.MetricData{types.MakeMetricData("foo==.baz",
-				[]float64{1, 2, 3, 4, 5}, 1, now32)},
-		},
+		// FIXME: I don't think `metric1.foo==.bar.baz` is a valid metric name, that is making this test fail https://github.com/grafana/carbonapi/issues/68
```
msaf1980, line 134: Equal not forbidden in plain metrics. For tags first ident `;` symbols used.

expr/functions/averageSeriesWithWildcards/function.go
```diff
@@ -71,7 +71,8 @@ func (f *averageSeriesWithWildcards) Do(ctx context.Context, e parser.Expr, from
 	for i, series := range nodeList {
 		args := groups[series]
 		name := "averageSeriesWithWildcards(" + series + ")"
-		r := args[0].CopyTag(name, map[string]string{"name": series})
+		r := args[0].CopyTag(name, map[string]string{"name": name})
```
msaf1980, line 74: Name tag must be cleaned from function artifacts. If not, alias functions need extract metric name from them. Thats really needed ?

expr/functions/cairo/png/cairo.go
```diff
@@ -789,10 +790,13 @@ func EvalExprGraph(ctx context.Context, e parser.Expr, from, until int64, values
 					return nil, err
 				}
 				r.Dashed = d
+				r.Tags["dashed"] = fmt.Sprintf("%f", d)
```
msaf1980, line 793: Spinntf nas bad perfomance. May be replace with strconv.FormatFloat ?

expr/functions/delay/function.go
```diff
@@ -62,8 +63,9 @@ func (f *delay) Do(ctx context.Context, e parser.Expr, from, until int64, values
 			prevValues = append(prevValues, value)
 		}
 
-		result := series.CopyTag("delay("+series.Name+","+stepsStr+")", series.Tags)
+		result := series.CopyName("delay(" + series.Name + "," + stepsStr + ")")
 		result.Values = newValues
+		result.Tags["delay"] = fmt.Sprintf("%d", steps)
```
msaf1980, line 68: May be simple use stepsStr or if really needed, convert to string outside a cycle for better perfomance ?

expr/functions/divideSeries/function.go
```diff
@@ -78,7 +80,7 @@ func (f *divideSeries) Do(ctx context.Context, e parser.Expr, from, until int64,
 		return nil, errors.New("must be called with 2 series or a wildcard that matches exactly 2 series")
 	}
 
-	alignedSeries := helper.AlignSeries(append(numerators, denominator))
+	alignedSeries := helper.AlignSeries(types.CopyMetricDataSlice(append(numerators, denominator)))
```
msaf1980, line 83: Possible two-time copy (append, and CopyMetricDataSlice). May be one pass alloc, append from nummunators, append from denominator ?

expr/functions/divideSeries/function_test.go
```diff
@@ -131,11 +131,7 @@ func TestDivideSeriesAligned(t *testing.T) {
 	for _, tt := range tests {
 		testName := tt.Target
 		t.Run(testName, func(t *testing.T) {
-			err := th.TestEvalExprModifiedOrigin(t, &tt, 0, 1, false)
-			if err != nil {
-				t.Errorf("unexpected error while evaluating %s: got `%+v`", tt.Target, err)
-				return
-			}
+			th.TestEvalExpr(t, &tt)
```
msaf1980, line 134: Why remove ?  It's for checks missed copy-on-write.

expr/functions/exp/function.go
```diff
@@ -0,0 +1,75 @@
...
+func (f *exp) Do(ctx context.Context, e parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types.MetricData, error) {
+	args, err := helper.GetSeriesArg(ctx, e.Args()[0], from, until, values)
+	if err != nil {
+		return nil, err
+	}
+	var results []*types.MetricData
+
+	for _, a := range args {
+		r := a.CopyLink()
+		r.Name = fmt.Sprintf("exp(%s)", a.Name)
```
msaf1980, line 41: May befor better perfomance
r.Name = "exp(" + a.Name + ")"

expr/functions/exponentialMovingAverage/function.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	if len(arg) == 0 {
+		return results, nil
+	}
+
+	constant := float64(2 / (float64(windowSize) + 1))
+
+	for _, a := range arg {
+		r := a.CopyLink()
+		r.Name = fmt.Sprintf("%s(%s,%s)", e.Target(), a.Name, argstr)
```
msaf1980, line 82: Also mayby without dynamic Sprintf (which need scan format string and uses high count interfaces boxing/unboxing) ?

expr/functions/groupByNode/function_test.go
```diff
@@ -50,6 +50,18 @@ func TestGroupByNode(t *testing.T) {
 				"qux": {types.MakeMetricData("qux", []float64{13, 15, 17, 19, 21}, 1, now32)},
 			},
 		},
+		{
+			Name:   "groupByNode_with_tag",
```
msaf1980, line 54: Why this name ? Where the tags ?

expr/functions/groupByNode/function_test.go
```diff
@@ -115,35 +143,62 @@ func TestGroupByNode(t *testing.T) {
...
 			},
-			Name: "groupByNode_tagged_names_with_special_symbol_equal",
 			Results: map[string][]*types.MetricData{
-				"lag":  {types.MakeMetricData("lag", []float64{1, 2, 3, 4, 5}, 1, now32)},
-				"lag=": {types.MakeMetricData("lag=", []float64{1, 0, 3, 4, 5}, 1, now32)},
+				"metric1.foo.bar1.baz.foo": {types.MakeMetricData("metric1.foo.bar1.baz.foo", []float64{1, 2, 3, 4, 5}, 1, now32)},
+				"metric1.foo.bar1.bla.foo": {types.MakeMetricData("metric1.foo.bar1.bla.foo", []float64{1, 2, 3, 4, 5}, 1, now32)},
 			},
 		},
+		// FIXME: I don't think `..foo==.bar` is a valid metric name, that is making these tests fail https://github.com/grafana/carbonapi/issues/68
```
msaf1980, line 171: It's not forbidden symbols. Not used for first tags ident.

expr/functions/identity/function.go
```diff
@@ -0,0 +1,81 @@
...
+	newValues := make([]float64, (until-from-1+step)/step)
+	value := from
+	for i := 0; i < len(newValues); i++ {
+		newValues[i] = float64(value)
+		value += step
+	}
+
+	p := types.MetricData{
+		FetchResponse: pb.FetchResponse{
+			Name:              fmt.Sprintf("identity(%s)", name),
```
msaf1980, line 49: Mayy be without Sprintf ?

expr/functions/integralByInterval/function.go
```diff
@@ -44,28 +44,25 @@ func (f *integralByInterval) Do(ctx context.Context, e parser.Expr, from, until
...
-				StepTime:          arg.StepTime,
-				StartTime:         arg.StartTime,
-				StopTime:          arg.StopTime,
-				XFilesFactor:      arg.XFilesFactor,
-				PathExpression:    name,
-				ConsolidationFunc: arg.ConsolidationFunc,
-			},
-			Tags: arg.Tags,
-		}
+		name := fmt.Sprintf("integralByInterval(%s,'%s')", arg.Name, e.Args()[1].StringValue())
```
msaf1980, line 58: May be without Sprintf

expr/functions/integralByInterval/function.go
```diff
@@ -44,28 +44,25 @@ func (f *integralByInterval) Do(ctx context.Context, e parser.Expr, from, until
...
+		result := arg.CopyLink()
+		result.Name = name
+		result.PathExpression = name
+		result.Values = make([]float64, len(arg.Values))
+
+		result.Tags["integralByInterval"] = intervalString
```
msaf1980, line 64: Tags need be copied before append new tags or modify existing

expr/functions/linearRegression/function.go
```diff
@@ -42,11 +43,11 @@ func (f *linearRegression) Do(ctx context.Context, e parser.Expr, from, until in
...
 
 	for _, a := range arg {
-		r := *a
-		if e.ArgsLen() > 2 {
-			r.Name = "linearRegression(" + a.Name + ",'" + e.Arg(1).StringValue() + "','" + e.Arg(2).StringValue() + "')"
-		} else if e.ArgsLen() > 1 {
-			r.Name = "linearRegression(" + a.Name + ",'" + e.Arg(1).StringValue() + "')"
+		r := a.CopyLink()
+		if len(e.Args()) > 2 {
+			r.Name = fmt.Sprintf("linearRegression(%s,'%s','%s')", a.GetName(), e.Arg(1).StringValue(), e.Arg(2).StringValue())
```
msaf1980, line 48: May be without Sprintf ?

expr/functions/linearRegression/function.go
```diff
@@ -87,7 +88,9 @@ func (f *linearRegression) Do(ctx context.Context, e parser.Expr, from, until in
 		for i := range r.Values {
 			r.Values[i] = consolidations.Poly(float64(i), c.RawMatrix().Data...)
 		}
-		results = append(results, &r)
+		r.Tags["linearRegressions"] = fmt.Sprintf("%d, %d", a.GetStartTime(), a.GetStopTime())
```
msaf1980, line 91: Aso, may be without Sprintf ?

expr/functions/moving/function.go
```diff
@@ -107,6 +110,41 @@ func (f *moving) Do(ctx context.Context, e parser.Expr, from, until int64, value
...
+		}
+	} else if len(e.Args()) == 3 {
+		xFilesFactor, err = e.GetFloatArgDefault(2, float64(arg[0].XFilesFactor))
+
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	switch e.Target() {
```
msaf1980, line 138: May be better use numbered enum ?

expr/functions/movingMedian/function.go
```diff
@@ -144,6 +144,7 @@ func (f *movingMedian) Do(ctx context.Context, e parser.Expr, from, until int64,
 				}
 			}
 		}
+		r.Tags["movingMedian"] = fmt.Sprintf("%d", windowSize)
```
msaf1980, line 147: May be without Sprintf ?

expr/functions/multiplySeriesWithWildcards/function.go
```diff
@@ -69,10 +69,24 @@ func (f *multiplySeriesWithWildcards) Do(ctx context.Context, e parser.Expr, fro
...
 	results := make([]*types.MetricData, 0, len(nodeList))
+	commonTags := helper.GetCommonTags(args)
+	commonTags["aggregatedBy"] = "multiply"
 
 	for _, series := range nodeList {
 		args := groups[series]
-		r := args[0].CopyTag("multiplySeriesWithWildcards("+series+")", map[string]string{"name": series})
+		r := args[0].CopyLink()
+		r.Name = series
+		r.Tags = make(map[string]string)
```
msaf1980, line 79: May be incapsylate logic for tags copy in additional Copy function ? Many cases in this PR need them for simplify logic.

msaf1980, line 79: May be encapsulate logic for tags copy in additional Copy function ? Many cases in this PR need them for simplify logic.

expr/functions/nPercentile/function.go
```diff
@@ -44,10 +45,10 @@ func (f *nPercentile) Do(ctx context.Context, e parser.Expr, from, until int64,
 
 	results := make([]*types.MetricData, len(arg))
 	for i, a := range arg {
-		r := *a
-		r.Name = "nPercentile(" + a.Name + "," + percentStr + ")"
+		r := a.CopyLink()
+		r.Name = fmt.Sprintf("nPercentile(%s,%s)", a.Name, percentStr)
```
msaf1980, line 49: May be without Sprintf ?

expr/functions/nPercentile/function.go
```diff
@@ -44,10 +45,10 @@ func (f *nPercentile) Do(ctx context.Context, e parser.Expr, from, until int64,
...
 		r.Values = make([]float64, len(a.Values))
-
+		r.Tags["nPercentile"] = fmt.Sprintf("%f", percent)
```
msaf1980, line 51: May be reuse already buffered to string conversion
r.Tags["nPercentile"] = percentStr

expr/functions/nonNegativeDerivative/function.go
```diff
@@ -85,9 +85,10 @@ func (f *nonNegativeDerivative) Do(ctx context.Context, e parser.Expr, from, unt
 			name = "nonNegativeDerivative(" + a.Name + ")"
 		}
 
-		r := *a
+		r := a.CopyLink()
 		r.Name = name
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["nonNegativeDerivative"] = "1"
```
msaf1980, line 91: Need copy-on-write for tags. May be encapsylate logic with special Copy function ?

expr/functions/offset/function.go
```diff
@@ -39,17 +40,18 @@ func (f *offset) Do(ctx context.Context, e parser.Expr, from, until int64, value
 		return nil, err
 	}
 	factorStr := strconv.FormatFloat(factor, 'g', -1, 64)
-
 	results := make([]*types.MetricData, len(arg))
+
 	for i, a := range arg {
-		r := *a
-		r.Name = e.Target() + "(" + a.Name + "," + factorStr + ")"
+		r := a.CopyLink()
```
msaf1980, line 46: Also need copy-on write for tags. And may be without Sprintf ?

expr/functions/offsetToZero/function.go
```diff
@@ -41,6 +42,7 @@ func (f *offsetToZero) Do(ctx context.Context, e parser.Expr, from, until int64,
 		for i, v := range a.Values {
 			r.Values[i] = v - minimum
 		}
+		r.Tags["offsetToZero"] = fmt.Sprintf("%f", minimum)
```
msaf1980, line 45: May be without Spirntf ?

expr/functions/pow/function.go
```diff
@@ -40,12 +41,13 @@ func (f *pow) Do(ctx context.Context, e parser.Expr, from, until int64, values m
 		return nil, err
 	}
 	factorStr := strconv.FormatFloat(factor, 'g', -1, 64)
-
 	results := make([]*types.MetricData, len(arg))
+
 	for j, a := range arg {
-		r := *a
-		r.Name = "pow(" + a.Name + "," + factorStr + ")"
+		r := a.CopyLink()
```
msaf1980, line 47: Also, need copy-on-write tags. And may be without Sprintf ?

expr/functions/pow/function.go
```diff
@@ -40,12 +41,13 @@ func (f *pow) Do(ctx context.Context, e parser.Expr, from, until int64, values m
...
+		r.Name = fmt.Sprintf("pow(%s,%s)", a.Name, factorStr)
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["pow"] = fmt.Sprintf("%f", factor)
```
msaf1980, line 50: May be 
r.Tags["pow"] = factorStr

expr/functions/rangeOfSeries/function.go
```diff
@@ -34,10 +34,20 @@ func (f *rangeOfSeries) Do(ctx context.Context, e parser.Expr, from, until int64
...
 		return nil, err
 	}
+	if len(series) == 0 {
+		return []*types.MetricData{}, nil
+	}
 
 	r := series[0].CopyName(e.Target() + "(" + e.RawArgs() + ")")
 	r.Values = make([]float64, len(series[0].Values))
 
+	commonTags := helper.GetCommonTags(series)
```
msaf1980, line 44: May be double copy of tags (in CopyName and in GetCommonTags) not needed ?

expr/functions/removeBetweenPercentile/function.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		if len(values) > 0 {
+			lowerThresholds = append(lowerThresholds, consolidations.Percentile(values, (100.0-percentile), false))
+			higherThresholds = append(higherThresholds, consolidations.Percentile(values, percentile, false))
+		}
+	}
+
+	for i, a := range args {
+		r := a.CopyLink()
+		r.Name = fmt.Sprintf("%s(%s, %g)", e.Target(), a.Name, number)
```
msaf1980, line 72: May be without Sprintf, number can be converted to string outside a cycle. But not required, may be need benchmark for see costs of them.

expr/functions/removeEmptySeries/function.go
```diff
@@ -55,8 +60,11 @@ func (f *removeEmptySeries) Do(ctx context.Context, e parser.Expr, from, until i
 				}
 			}
 		}
-		if nonNull != 0 && nonNull/float64(len(arg.Values)) >= factor {
-			results = append(results, arg)
+
+		if nonNull != 0 && helper.XFilesFactor(nonNull, len(arg.Values), xFilesFactor) {
```
msaf1980, line 64: May be add some tests for XFilesFactor ?

expr/functions/round/function.go
```diff
@@ -56,19 +56,14 @@ func (f *round) Do(ctx context.Context, e parser.Expr, from, until int64, values
...
 		}
-		results[j] = &r
-	}
-	return results, nil
-}
 
-func doRound(x float64, precision int) float64 {
-	if math.IsNaN(x) {
-		return x
+		r.Tags["round"] = fmt.Sprintf("%d", precision)
```
msaf1980, line 62: We really need string conversion in each call of function ? We have preconverted precisionStr.

expr/functions/scale/function.go
```diff
@@ -47,14 +47,15 @@ func (f *scale) Do(ctx context.Context, e parser.Expr, from, until int64, values
...
-		r := *a
+		r := a.CopyLink()
 		if timestamp == 0 {
 			r.Name = "scale(" + a.Name + "," + scaleStr + ")"
 		} else {
 			r.Name = fmt.Sprintf("scale(%s,%g,%d)", a.Name, scale, timestamp)
 			r.Name = "scale(" + a.Name + "," + scaleStr + "," + e.Arg(2).StringValue() + ")"
 		}
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["scale"] = fmt.Sprintf("%f", scale)
```
msaf1980, line 58: May be
r.Tags["scale"] = scaleStr

expr/functions/scaleToSeconds/function.go
```diff
@@ -43,16 +44,18 @@ func (f *scaleToSeconds) Do(ctx context.Context, e parser.Expr, from, until int6
 	results := make([]*types.MetricData, len(arg))
 
 	for j, a := range arg {
-		r := *a
+		r := a.CopyLink()
 		r.Name = "scaleToSeconds(" + a.Name + "," + secondsStr + ")"
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["scaleToSeconds"] = fmt.Sprintf("%f", seconds)
```
msaf1980, line 50: Also, may be
r.Tags["scaleToSeconds"] = secondsStr

expr/functions/sigmoid/function.go
```diff
@@ -0,0 +1,76 @@
...
+func (f *sigmoid) Do(ctx context.Context, e parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types.MetricData, error) {
+	arg, err := helper.GetSeriesArg(ctx, e.Args()[0], from, until, values)
+	if err != nil {
+		return nil, err
+	}
+	var results []*types.MetricData
+
+	for _, a := range arg {
+		r := a.CopyLink()
+		r.Name = fmt.Sprintf("sigmoid(%s)", a.Name)
```
msaf1980, line 42: May be without Snprinf ?

expr/functions/timeSlice/function.go
```diff
@@ -56,9 +57,11 @@ func (f *timeSlice) Do(ctx context.Context, e parser.Expr, from, until int64, va
 	results := make([]*types.MetricData, len(arg))
 
 	for n, a := range arg {
-		r := *a
+		r := a.CopyLink()
 		r.Name = "timeSlice(" + a.Name + "," + startStr + "," + endStr + ")"
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["timeSliceStart"] = fmt.Sprintf("%d", start)
```
msaf1980, line 63: May be use startStr and endStr for avoid a conversion in cycle ?

expr/functions/timeStack/function.go
```diff
@@ -56,11 +56,13 @@ func (f *timeStack) Do(ctx context.Context, e parser.Expr, from, until int64, va
 		}
 
 		for _, a := range arg {
-			r := *a
+			r := a.CopyLink()
 			r.Name = fmt.Sprintf("timeShift(%s,%d)", a.Name, offs)
```
msaf1980, line 60: Also, may be avoid Spprintf and use preconverted offsStr ?

expr/functions/timeStack/function.go
```diff
@@ -56,11 +56,13 @@ func (f *timeStack) Do(ctx context.Context, e parser.Expr, from, until int64, va
...
 			r.StartTime = a.StartTime - offs
 			r.StopTime = a.StopTime - offs
-			results = append(results, &r)
+			r.Tags["timeShiftUnit"] = fmt.Sprintf("%d", unit)
```
msaf1980, line 63: May be use preconverted unitStr ?

expr/functions/timeStack/function.go
```diff
@@ -56,11 +56,13 @@ func (f *timeStack) Do(ctx context.Context, e parser.Expr, from, until int64, va
...
+			r.Tags["timeShift"] = fmt.Sprintf("%d", offs)
```
msaf1980, line 64: May be use strconv.FormatInt ?

expr/functions/transformNull/function.go
```diff
@@ -91,8 +91,9 @@ func (f *transformNull) Do(ctx context.Context, e parser.Expr, from, until int64
 			name = "transformNull(" + a.Name + ")"
 		}
 
-		r := a.CopyTag(name, a.Tags)
+		r := a.CopyName(name)
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["transformNull"] = fmt.Sprintf("%f", defv)
```
msaf1980, line 96: May be
r.Tags["transformNull"] = defvStr

expr/helper/helper.go
```diff
@@ -93,7 +93,8 @@ func GetSeriesArgsAndRemoveNonExisting(ctx context.Context, e parser.Expr, from,
 func AggKey(arg *types.MetricData, nodesOrTags []parser.NodeOrTag) string {
 	matched := make([]string, 0, len(nodesOrTags))
 	metricTags := arg.Tags
-	nodes := strings.Split(metricTags["name"], ".")
+	name := metric.ExtractMetric(arg.Name)
```
msaf1980, line 96: Name already in tag name ? Why we need extract them again ?

expr/helper/helper.go
```diff
@@ -161,7 +143,7 @@ func ForEachSeriesDo(ctx context.Context, e parser.Expr, from, until int64, valu
 	var results []*types.MetricData
 
 	for _, a := range arg {
-		r := a.CopyTag(e.Target()+"("+a.Name+")", a.Tags)
+		r := a.CopyName(e.Target() + "(" + a.Name + ")")
```
msaf1980, line 146: CopyName do copy of tags. We really need them ? Where tags are ovewtitten ?

expr/types/types.go
```diff
@@ -448,35 +448,13 @@ func (r *MetricData) CopyLinkTags() *MetricData {
...
 }
 
-// CopyName returns the copy of MetricData, Values not copied and link from parent. If name set, Name and Name tag changed, Tags wil be reset
+// CopyName returns the copy of MetricData, Values not copied and link from parent. If name set, Name and Name tag changed
 func (r *MetricData) CopyName(name string) *MetricData {
-	if name == "" {
-		return r.CopyLink()
-	}
+	res := r.CopyLink()
+	res.Name = name
```
msaf1980, line 454: No nneded. CopyLink do this internally.

tests/helper.go
```diff
@@ -264,6 +264,8 @@ func TestMultiReturnEvalExpr(t *testing.T, tt *MultiReturnEvalTestItem) {
 		}
 
 		for k, v := range wants[0].Tags {
+			// FIXME: we are skipping Tags comparison until we fix the tests https://github.com/grafana/carbonapi/issues/65
```
msaf1980, line 267: Need to be fixed. It will mask overwrite tags without copy-on-write. And produce some bugs. Lookup the failed tests of functions and and copy-onwrite when needed.

tests/helper.go
```diff
@@ -274,6 +276,8 @@ func TestMultiReturnEvalExpr(t *testing.T, tt *MultiReturnEvalTestItem) {
 		}
 
 		for k := range actual.Tags {
+			// FIXME: we are skipping Tags comparison until we fix the tests https://github.com/grafana/carbonapi/issues/65
+			break
```
msaf1980, line 280: Need to be fixed. It will mask overwrite tags without copy-on-write. And produce some bugs. Lookup the failed tests of functions and and copy-onwrite when needed.

tests/helper.go
```diff
@@ -401,6 +405,8 @@ func TestEvalExprModifiedOrigin(t *testing.T, tt *EvalTestItem, from, until int6
 			t.Errorf("metric[%d] %+v with name %v doesn't contain 'name' tag", i, actual, actual.Name)
 		}
 		for k, v := range want.Tags {
+			// FIXME: we are skipping Tags comparison until we fix the tests https://github.com/grafana/carbonapi/issues/65
```
msaf1980, line 408: Need to be fixed. It will mask overwrite tags without copy-on-write. And produce some bugs. Lookup the failed tests of functions and and copy-onwrite when needed.

tests/helper.go
```diff
@@ -410,6 +416,8 @@ func TestEvalExprModifiedOrigin(t *testing.T, tt *EvalTestItem, from, until int6
 			}
 		}
 		for k := range actual.Tags {
+			// FIXME: we are skipping Tags comparison until we fix the tests https://github.com/grafana/carbonapi/issues/65
```
msaf1980, line 419: Need to be fixed. It will mask overwrite tags without copy-on-write. And produce some bugs. Lookup the failed tests of functions and and copy-onwrite when needed.

tests/helper.go
Civil: So the PR they are refering to here says:
```
Solving this in the upstream sync PR would have made the PR huge. So I'm creating this issue to solve it separately, and we will be skipping the Tags comparison for now.
```

I would say this is for now valid, and can indeed be fixed in a separate PR later (this is already a big set of change and I would agree that fixing that in the same PR would make it much harder to review).

expr/functions/aliasByBase64/function.go
npazosmendez: Looks like this change was a mistake on our side, since `CopyName` also updates the `"name"` tag, and that is correct. Will fix.

expr/functions/aliasByMetric/function.go
npazosmendez: We noticed the previous version was failing for cases where the nested result was named after the function that processed it. For example:

`aliasByMetric(ifft(foo.bar))`'


In that case, the argument of `aliasByMetric` is a series with name `ifft(foo.bar)`, which is not a metric name and the `strings.Split(metric, ".")` gives wrong results: `ifft(foo`, `bar)`. So we fixed it this was, by correctly extracting the metric name. Let us know if there's a better way.

expr/functions/averageSeriesWithWildcards/function.go
npazosmendez: That's interesting, since it would explain the fix we had to do related to `ExtractMetric`. But we've been taking GraphiteWeb as a reference, and it does seem to include functions in the `name` tag; for example:

https://github.com/graphite-project/graphite-web/blob/ed08fa4e2f6526ae11e55f88f70d0b6d36e4052f/webapp/graphite/render/functions.py#L243-L249

And from what I see, functions like `aliasByNode` are indeed extracting the name from them: https://github.com/graphite-project/graphite-web/blob/ed08fa4e2f6526ae11e55f88f70d0b6d36e4052f/webapp/graphite/render/functions.py#L111

Please let us know if you think we are misunderstanding something.

expr/functions/groupByNode/function_test.go
npazosmendez: This test is about `groupByNode` arguments referring to tags rather than node indexes. See below:

```
groupByNode(metric1.foo.*.*,"name","sum")
```

That is telling `groupByNode` to group all the metrics with the same "name" tag (for this test, only one). From the [docs](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.groupByNode) 

> Node may be an integer referencing a node in the series name or a string identifying a tag.

expr/functions/integralByInterval/function.go
npazosmendez: I think they are being copied, look at the `CopyLink` call a few lines above:


https://github.com/grafana/carbonapi/blob/d511d4c9150c851de8e77cd4f3014071ad754e61/expr/functions/integralByInterval/function.go#L59-L64

expr/functions/averageSeriesWithWildcards/function.go
msaf1980: It's not a simple question. ExtractName can extract name/tags from them. But why it's really needed ?
And how to resolve aliasByTags(series, 'name)

With this PR, returned ```averageSeriesWithWildcards... ``` And need some additional strip. May be do it in place, as are in head ? 100% compability of internal structs usually not required. What are changed this for display in graphana dashboard ? But it's disputable.

expr/functions/integralByInterval/function.go
msaf1980: Yes, you are right.

expr/functions/perSecond/function.go
```diff
@@ -87,9 +87,10 @@ func (f *perSecond) Do(ctx context.Context, e parser.Expr, from, until int64, va
 			name = "perSecond(" + a.Name + ")"
 		}
 
-		r := a.CopyLinkTags()
+		r := a.CopyLink()
 		r.Name = name
 		r.Values = make([]float64, len(a.Values))
+		r.Tags["perSecond"] = "1"
```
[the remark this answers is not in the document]
npazosmendez: `r := a.CopyLink()` is called a few lines above, so I think it should be fine

expr/functions/linearRegression/function.go
npazosmendez: Sounds good. I'll try to address this and all your other comments related to Sprintf

expr/functions/nonNegativeDerivative/function.go
Civil: Just a thought for some distant bright future: I wanted to use semgrep to automatically check that tags are copied when needed. AFAIR it should be able to warn when some one is not using CopyLink but modifying tags. Maybe now it would be worth investing time into that.

expr/functions/averageSeriesWithWildcards/function.go
npazosmendez: >  100% compability of internal structs usually not required. What are changed this for display in graphana dashboard ? 

Well, the `"name"` tag is not something internal, so setting it to `foobar` vs setting it to `averageSeriesWithWildcards(foobar)` does make a difference. And I think it's not only about visualization, it can lead to different results when combined with other functions; for example  `grep(aliasByNode(averageSeries(...), "name"), "average")`.

Do you have a reference to GraphiteWeb's docs or code that suggest that the `"name"` tag should always be set to the metric name without function names?

msaf1980: aliasByNodes/aliasByTags functions will work corect with this, ExtractMetric is used. But may be in other.
https://github.com/grafana/carbonapi/blob/d511d4c9150c851de8e77cd4f3014071ad754e61/expr/types/types.go#L468

But you are right. Some functions like can produce a problem.
Like `grep(aliasByNode(averageSeries(...), "name"), "average").`
It's disputable question, in my opinion search on name without function is better.
But if in graphite-web realize different logic -  so, let it be. No objections.
May be more tests for correct tags extraction is needed, but now I can't say in which functions pipeline it's needed.

PS. Some functions in graphite API is dangerous, like `aliasSub`, which applied to raw name (with functions and args). and can corrupt name.
But capability :-(

expr/functions/nonNegativeDerivative/function.go
msaf1980: It's my mistake, CopyLink copy tags map. Perhaps I chose not very good names.

expr/functions/offset/function.go
msaf1980: Your are right. It's hard to keep your attention on with such changes.

tests/helper.go
msaf1980: > I would say this is for now valid, and can indeed be fixed in a separate PR later (this is already a big set of change and I would agree that fixing that in the same PR would make it much harder to review).
May be, but need don't forget about this. No-Copy-On-Write bugs is my nightmare :-) I spent quite a lot of time fixing a some  of bugs in huge functions pipeline.

expr/functions/averageSeriesWithWildcards/function.go
Civil: In that case I think solution is to create internal tag (e.x. `__name__` or something like that) that wouldn't be visible to the user or just internal field in MetricData struct. As one of the problems here is if you have full UTF-8 support enabled and all symbols are allowed in a metric name, you will have problems parsing the result, and while on the parser level you have an idea what is metric name and what is not, but on a function level ExtractMetric will produce malformed results sometimes because at that level we don't have the whole syntax tree, but rather just a string with text.

expr/functions/removeEmptySeries/function.go
carrieedwards: It looks like there are a few tests using XFilesFactor for removeEmptySeries and one for removeZeroSeries. Would you like us to add a few more tests?

tests/helper.go
npazosmendez: We won't forget. But to clarify: this is not about copy-on-write, the assertions on the original data being unmodified are still there (and tests are passing).

This is commenting out the assertions on the **tags of the output**. And we are commenting them out not because the functions implementations are wrong, but because the tests are not currently building the "expected" MetricData correctly.

expr/helper/helper.go
npazosmendez: There are the functions that call `ForEachSeriesDo`:
- `absolute`
- `derivative`
- `integral`
- `invert`
- `isNotNull`
- `offsetToZero`

And they all seem to modify the tags. I see your point though: a function calling `ForEachSeriesDo` might not necessarily want to change the tags, so this mightbe unnecessary overhead for a future use case. Maybe we could have the functions I enumerated above copy the tags manually as they need, but I think this is not necessary now. Let me know what you think,

expr/functions/cairo/png/cairo.go
```diff
@@ -790,7 +791,7 @@ func EvalExprGraph(ctx context.Context, e parser.Expr, from, until int64, values
 					return nil, err
 				}
 				r.Dashed = d
-				r.Tags["dashed"] = fmt.Sprintf("%f", d)
+				r.Tags["dashed"] = strconv.FormatFloat(d, 'g', -1, 64)
```
msaf1980, line 794: Move this (and conversion also) out of cycle. Also replace `r := *a` with CopyLink() for prevent no-copy-on write bug.

expr/functions/integralByInterval/function.go
```diff
@@ -55,7 +54,7 @@ func (f *integralByInterval) Do(ctx context.Context, e parser.Expr, from, until
 		current := 0.0
 		currentTime := arg.StartTime
 
-		name := fmt.Sprintf("integralByInterval(%s,'%s')", arg.Name, e.Args()[1].StringValue())
+		name := "integralByInterval(" + arg.Name + ",'" + e.Args()[1].StringValue() + "')"
```
msaf1980, line 57: Replace `e.Args()[1]` with Arg(1) and move outside a cycle. It's old artifact (not cleaned). See
https://github.com/grafana/carbonapi/blob/9eb0a9e50888933515bb9756dbdd8954db423cf1/pkg/parser/parser.go#L111
Like

argStr := e.Arg(1).StringValue()
for j, arg := range args {
		name := "integralByInterval(" + arg.Name + ",'" + argStr + "')"

It's save some allocations during interface unboxing.

expr/functions/linearRegression/function.go
```diff
@@ -45,9 +45,9 @@ func (f *linearRegression) Do(ctx context.Context, e parser.Expr, from, until in
 	for _, a := range arg {
 		r := a.CopyLink()
 		if len(e.Args()) > 2 {
-			r.Name = fmt.Sprintf("linearRegression(%s,'%s','%s')", a.GetName(), e.Arg(1).StringValue(), e.Arg(2).StringValue())
+			r.Name = "linearRegression(" + a.GetName() + ",'" + e.Arg(1).StringValue() + "','" + e.Arg(2).StringValue() + "')"
```
msaf1980, line 48: Also move e.Arg(N).StringValue() outside a cycle/. But not blocking change, may be for simplify PR refactoing we can merge without this.

expr/functions/removeEmptySeries/function.go
msaf1980: May be not now. It's not blocking.

expr/helper/helper.go
msaf1980: Yes, you are right. May be rename ForEachSeriesDo and ForEachSeriesDo1 with more adequate name. But not in this PR.

tests/helper.go
msaf1980: Yes, but some protection and test correction is needed. But may be not in this PR.

expr/functions/cairo/png/cairo.go
npazosmendez: Done. Note that I had to define `dashLen` and `dashLenStr` that are only used in the `"dashed"` case.  See https://github.com/go-graphite/carbonapi/pull/686/commits/1ecf628951964ebe99c77dd495be865304e6fb09

expr/functions/integralByInterval/function.go
npazosmendez: Interesting. Thank you for the insight. I just noticed `intervalString` is defined in line 46, so I'll use that for this case.

tests/helper.go
npazosmendez: Will send this in another PR once this is merged. I'll resolve this now, feel free to re-open if you disagree.

expr/functions/divideSeries/function.go
npazosmendez: I think it should be fine because it's an array of pointers `[]*types.MetricData`, so it's only `CopyMetricDataSlice([]*MetricData)` copying the actual MetricData. I don't see a way of avoiding a potential copy of the pointers when appending. Let me know if you think it's OK or if I misunderstood you.

msaf1980: Not problem, this is not blocking. Cost of copying slice of pointers not too high. May be I'll add some benchmarks later and change them, if needed.

expr/functions/aliasByNode/function_test.go
npazosmendez: You are right. Then I think there's a bug in the `metric.ExtractMetric` helper, because if I test this it fails (here 
https://github.com/go-graphite/carbonapi/blob/b1e09e902ff9b2fd2bc9c8f0ec16d95fe885575d/expr/helper/metric/extract_test.go#L5):
```go
		{
			"metric1.foo.Ab1==.lag",
			"metric1.foo.Ab1==.lag",
		},
```
```
--- FAIL: TestExtractMetric (0.00s)
    --- FAIL: TestExtractMetric/metric1.foo.Ab1==.lag (0.00s)
        /home/nicolas/code/carbonapi/expr/helper/metric/extract_test.go:51: extractMetric("metric1.foo.Ab1==.lag")=".lag", want "metric1.foo.Ab1==.lag"
FAIL
FAIL	github.com/go-graphite/carbonapi/expr/helper/metric	0.002s
FAIL
```

Civil (approved):

expr/functions/aliasByNode/function_test.go
msaf1980: Yes, helper.ExtractMetric has this bug. For extatract nodes I use types.ExtractName, but need some checks
https://github.com/go-graphite/carbonapi/blob/main/expr/types/extract.go#L63

msaf1980: But it need some adpotation for this

msaf1980: I add some fix.
https://github.com/msaf1980/carbonapi/commit/0f10311b44ce040d7f6523fc526b309c8201909c

But now we need fix for all functions. `types.MakeMetricData` don't get manual-added tags (derived from graphite-web).
For example, fixes for transformNull
https://github.com/msaf1980/carbonapi/commit/1e042b829936f00a50b91d088f3447e974726e98

npazosmendez: > I add some fix.
https://github.com/msaf1980/carbonapi/commit/0f10311b44ce040d7f6523fc526b309c8201909c

Nice, that works perfectly. I'll incorporate the `ExtractNameTag`.

> But now we need fix for all functions for tags check. 

That's another problem, the one we were discussing here: https://github.com/go-graphite/carbonapi/pull/686#discussion_r965673193 . I'm planning to fix all the tests using `types.MakeMetricData` in a separate PR so this one doesn't get bigger. This is the issue for that: https://github.com/grafana/carbonapi/issues/65

expr/functions/groupByNode/function_test.go
npazosmendez: Solved here [`d77e3ee` (#686)](https://github.com/go-graphite/carbonapi/pull/686/commits/d77e3ee963f7262ee852034f6fdcfa27663bca92)

expr/functions/averageSeriesWithWildcards/function.go
npazosmendez: I like the idea of the internal field. That would make the metric name extraction less error-prone and more efficient. For example, this call wouldn't be necessary: https://github.com/go-graphite/carbonapi/pull/686#discussion_r965554117 

But I think it would be better to leave this change out of this PR, I imagine it will be rather big. And I don't see any bugs related to this with the current state of the code anyway, so this PR shouldn't leave the code broken.

If you agree, I can create an issue so we can track it after this is merged.

expr/functions/divideSeries/function_test.go
npazosmendez: Resolving this, please feel free to re-open if needed.

expr/functions/averageSeriesWithWildcards/function.go
Civil: I would agree with that.

npazosmendez:
I think we are getting close to merging this. This is what is pending:
- Resolve a few remaining comments (these shouldn't be major)
- Include this fix once it's merged, we think it's important https://github.com/grafana/carbonapi/pull/107 
- Create the issues for the tasks we agreed will be left for later and not tackled in this PR. I will make sure these are clearly enumerated so we don't lose track.

expr/functions/aliasByMetric/function.go
npazosmendez: This is related to the other discussion we had. I've created this issue to address it later: https://github.com/go-graphite/carbonapi/issues/691

expr/functions/moving/function.go
npazosmendez: You mean for the `cons` variable?

expr/functions/rangeOfSeries/function.go
npazosmendez: Nice catch, done [`9faff07` (#686)](https://github.com/go-graphite/carbonapi/pull/686/commits/9faff0746b138b1872c5c2bd60af72fda3f97613)

msaf1980 (dismissed):

Civil:
Can you rebase your branch? As that's the only thing that preventing me from merging it.

npazosmendez:
Ok, I'll resolve the conflicts and rebase it. There's also the DeepSource checks failing, do you want us to fix those?

Civil:
Not really, there is a problem with DeepSource as it mostly complains about tests and vendored deps, while they should be excluded (according to the config).

Btw, for compatibility checks, do you have some sort of validation scripts/tools or you do the work manually?

carrieedwards:
> Btw, for compatibility checks with graphite-web, do you have some sort of validation scripts/tools or you do the work manually?

So far, we have been doing manual compatibility checks with Graphite web

npazosmendez:
@Civil done, rebasing was causing lots of conflicts because of how our change history is, so I merged the `main` branch instead. But that should unlock the merging

PR #693, opened 2022-09-14
Check validity of consolidation function in functions that use Summarizer

carrieedwards:
This PR addresses a bug in the consolidations.SummarizeValues function, in which the specified consolidation function was not checked to verify that it was a valid function. This would cause a panic to occur, because in consolidations.SummarizeValues, the default case in the switch statement involves a string split which assumes that the consolidation function is 'p50'-'p99.9'. If the specified consolidation function was not 'p50'-'p99.9', then the panic would occur on the string split. This bug was originally found in the sortBy() function, but any function that calls consolidations.Summarizer could experience the same panic.


expr/consolidations/consolidations.go
```diff
@@ -35,6 +40,18 @@ var ConsolidationToFunc = map[string]func([]float64) float64{
 
 var AvailableSummarizers = []string{"sum", "total", "avg", "average", "avg_zero", "max", "min", "last", "current", "range", "rangeOf", "median", "multiply", "diff", "count", "stddev"}
```
Civil, line 41: I think `first` should be also added to the list here as well as you've implemented it just few lines below.

Civil (changes requested):
I think there is just one small change missing (new consolidation function is missing from list of available summarizers), otherwise LGTM

expr/consolidations/consolidations.go
carrieedwards: Done

Issue #106, opened 2015-12-19
Support format=svg for Cairo backend 


dgryski:
Th is may require adding support to the bindings.

dgryski:
Blocked on https://github.com/martine/gocairo/issues/4

JaderDias:
@dgryski this issue has been unblocked

PR #694, opened 2022-09-14
Parse string input for params of type Integer or Float (#106)

carrieedwards:
This PR will change the float and int parsing functions to be able to parse string input. If the param type is a string, it will attempt to use strconv.ParseFloat or strconv.ParseInt and if the string can be converted to a float, that float value will be used. This will allow more leniency for users input, and be more consistent with Graphite web. Graphite web's parsing of floats and strings can be found [here](https://github.com/graphite-project/graphite-web/blob/b52987ac97f49dcfb401a21d4b92860cfcbcf074/webapp/graphite/functions/params.py).

### DIFF
--- a/cmd/mockbackend/testcases/pr560/pr560.yaml
+++ b/cmd/mockbackend/testcases/pr560/pr560.yaml
@@ -31,13 +31,6 @@ test:
             - endpoint: "http://127.0.0.1:8081"
               delay: 1
               type: "GET"
-              URL: "/render?target=constantLine('2222')"
-              expectedResponse:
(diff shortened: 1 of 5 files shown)


Civil (dismissed):

npazosmendez:
@Civil ha sorry, I just pushed to rebase and immediately dismissed your approval.

Civil (dismissed):

Civil:
```
This branch has conflicts that must be resolved
```

So I think you'll need to rebase again, sorry.

npazosmendez (approved):

PR #696, opened 2022-09-14
Prevent panic in removeEmptySeries if series is nil and xFilesFactor …

carrieedwards:
This PR fixes a panic that was occurring when removeEmptySeries or removeZeroSeries was called with a series that is nil as well as an xFilesFactor argument. The panic was occurring when the xFilesFactor argument was attempting to be accessed: args[0].XFilesFactor. If args[0] is nil, a runtime error: index out of range [0] with length 0. error is thrown. To prevent this, if the series argument has a length of 0, an empty series is returned. This can also occur is removeEmptySeries/removeZeroSeries has another function nested inside of it, and that function returns a nil series.


Civil (approved):

Civil:
LGTM, but github complains about conflicts in the branch...

PR #701, opened 2022-09-17
Add support for Graphite web sumSeriesLists function

carrieedwards:
This PR adds support for the Graphite web sumSeriesLists function, defined as:

```
sumSeriesLists(seriesListFirstPos, seriesListSecondPos)
Iterates over a two lists and subtracts series lists 2 through n from series 1 list1[0] to list2[0], list1[1] to list2[1] and so on. The lists will need to be the same length

Example:

&target=sumSeriesLists(mining.{carbon,graphite,diamond}.extracted,mining.{carbon,graphite,diamond}.shipped)
An example above would be the same as running [sumSeries](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.sumSeries) for each member of the list:

?target=sumSeries(mining.carbon.extracted,mining.carbon.shipped)
&target=sumSeries(mining.graphite.extracted,mining.graphite.shipped)

This un
&target=sumSeries(mining.diamond.extracted,mining.diamond.shipped)
This is an alias for [aggregateSeriesLists](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.aggregateSeriesLists) with aggregation sum.
```

The function definition can also be found [here](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.sumSeriesLists).

Tests were added to verify the functionality according to the Graphite web implementation and the function definition.

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -163,188 +163,189 @@ reverse: default value mismatch: got (empty), should be false |
 | timeSlice | endSliceAt: type mismatch: got interval, should be date |
 
 ## Supported functions
-| Function      | Carbonapi-only                                            |
-| :-------------|:--------------------------------------------------------- |
-| absolute(seriesList) | no |
-| add(seriesList, constant) | no |
-| aggregate(seriesList, func, xFilesFactor=None) | no |
-| aggregateLine(seriesList, func='average', keepStep=False) | no |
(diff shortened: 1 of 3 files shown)


npazosmendez (approved):

Civil (approved):

Civil:
Maybe for future changes, it would be worth to split functions and COMPATIBILITY.md change to avoid such blocker?

PR #704, opened 2022-09-21
Add support for Graphite web compressPeriodicGaps function

carrieedwards:
This PR adds support for the Graphite web compressPeriodicGaps function. This function is defined as:

```
compressPeriodicGaps(seriesList)
Tries to intelligently remove periodic None’s from series, recalculating start, stop and step values. You can use summarize(seriesList, ‘<desired step>’, ‘last’) function for that also, but this function trying to guess desired step automatically. Can be used in case of fix metric with improper resolution. Especially useful for derivative functions, which are not working with series with regular gaps.

Example:

&target=compressPeriodicGaps(Server.instance01.threads.busy)
```

Graphite web's implementation can be found [here](https://github.com/graphite-project/graphite-web/blob/b52987ac97f49dcfb401a21d4b92860cfcbcf074/webapp/graphite/render/functions.py#L5909). 

Tests matching those defined for this function in Graphite web have also been added.

### DIFF
--- a/expr/functions/compressPeriodicGaps/function.go
+++ b/expr/functions/compressPeriodicGaps/function.go
@@ -1,0 +1,140 @@
+package compressPeriodicGaps
+
+import (
+	"context"
+	"github.com/go-graphite/carbonapi/expr/consolidations"
+	"github.com/go-graphite/carbonapi/expr/helper"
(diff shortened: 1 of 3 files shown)


npazosmendez (approved):

Civil:
For new functions I think you can approve each others PRs and merge them. Or at least I hope that.

PR #706, opened 2022-09-21
Allow quoted numeric arguments for asPercent and keepLastValue

carrieedwards:
The asPercent function and the keepLastValue function were not allowing singly quoted input for their total and limit parameters respectively. Users will sometimes pass in numeric values within single quotes, and doing so was resulting in errors. Since we previously added ability to parse quoted floats and ints (see https://github.com/go-graphite/carbonapi/pull/694), it should be extended to asPercent and keepLastValue as well.


npazosmendez (approved):
LGTM

PR #710, opened 2022-09-22
Fix seriesList with incomplete arguments

msaf1980:
Also check arguments and return error without sort.

### DIFF
--- a/expr/functions/seriesList/function.go
+++ b/expr/functions/seriesList/function.go
@@ -43,7 +43,11 @@ func (f *seriesList) Do(ctx context.Context, e parser.Expr, from, until int64, v
 		return nil, err
 	}
 
-	numerators, err := helper.GetSeriesArg(ctx, e.Args()[0], from, until, values)
+	if e.ArgsLen() < 2 {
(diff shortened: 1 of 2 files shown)


expr/functions/seriesList/function.go
```diff
@@ -43,7 +43,11 @@ func (f *seriesList) Do(ctx context.Context, e parser.Expr, from, until int64, v
 		return nil, err
 	}
 
-	numerators, err := helper.GetSeriesArg(ctx, e.Args()[0], from, until, values)
+	if e.ArgsLen() < 2 {
+		return nil, types.ErrNeedMoreArguments
+	}
```
npazosmendez, line 48: This is already being done at the top of the function (though we used the `parser.ErrMissingArgument` error)

msaf1980: Oh, I see panic on older version. Fixed in main.

npazosmendez: Yeah it was a very recent fix https://github.com/go-graphite/carbonapi/pull/701

npazosmendez (approved):

PR #711, opened 2022-09-22
Update functions that were modifying original input

carrieedwards:
Several functions were still modifying the original series input, instead of making a copy before processing the data. This can lead to no-copy-on write bugs. This PR updates the remaining functions that were using `r := *a`, instead using `r := a.CopyLinkTags()`.


npazosmendez (approved):

msaf1980:
> `r := *a`
> `results[i] = &r`

No modify origin is this case. Color  is a  string
Main reson for use `CopyTags` and  avoid dereference `a` - copy to heap allocated again in original code - better performance.
CopyTags do linked copy in one step.

msaf1980 (approved):

PR #712, opened 2022-09-22
Handle default arguments and number of arguments to avoid errors

carrieedwards:
This PR adds checks for the number of arguments passed into a function, verifying that it meets the minimal number that the function can accept, and adds default arguments for functions that have those. This was based off of the function definitions defined in the [Graphite web documentation](https://graphite.readthedocs.io/en/latest/functions.html#). This should help prevent errors with indices, and properly handle default arguments in functions that have them.


Civil:
I sometimes think that we might work on some helper functions that will parse Description of a function and produce ready-to-consume arrays and maps for named arguments. But that never materialized beyond just fought.

PR #718, opened 2022-09-30
Fix ConsolidationFunc setting

npazosmendez:
We noticed some functions were setting the `r.AggregateFunction` directly instead of the string `r.ConsolidationFunc` that is then used in the `GetAggregateFunction` helper:

https://github.com/go-graphite/carbonapi/blob/30a090756e199c9d588499972268c4ebbb22a4e9/expr/types/types.go#L299-L310

This was leading to inconsistencies between `ConsolidationFunc` and `AggregateFunction` that were causing issues for our use case.


carrieedwards (approved):
Looks good to me

Civil (approved):

Civil (approved):

PR #721, opened 2022-10-05
Add time parsing that supports absolute and relative time

carrieedwards:
This PR expands CarbonAPI's handling of time values by parsing both absolute and relative time. This is needed to meet the requirements from the [Graphite web documentation](https://graphite.readthedocs.io/en/latest/render_api.html#from-until) as well as to match Graphite web's functionality (see [here](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/attime.py) for the implementation). 

This will fix time handling for several functions, include the time* functions, movingWindow, integral, and linearRegression.

### DIFF
--- a/pkg/parser/time_parser.go
+++ b/pkg/parser/time_parser.go
@@ -1,0 +1,241 @@
+package parser
+
+import (
+	"fmt"
+	"github.com/leekchan/timeutil"
+	"strconv"
+	"strings"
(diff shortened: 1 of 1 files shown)


Civil:
Hey, JFIY: I'm not looking at this PR because it's marked as a Draft.

PR #727, opened 2022-10-18
Fix smartSummarize alignTo interval handling to match Graphite web

carrieedwards:
The smartSummarize function is defined in the [Graphite web documentation](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.smartSummarize) as the following:

```
smartSummarize(seriesList, intervalString, func='sum', alignTo=None)
Smarter version of summarize.

The alignToFrom boolean parameter has been replaced by alignTo and no longer has any effect. **Alignment can be to years, months, weeks, days, hours, and minutes.**

This function can be used with aggregation functions average, median, sum, min, max, diff, stddev, count, range, multiply & last.
```

Note that the alignTo argument accepts values of "years", "months", "weeks", "days", "hours", and "minutes" (or their shortened version, such as "y", "m", "w", etc).  

Currently, CarbonAPI's version of smartSummarize uses the parser.GetIntervalString function, which requires an integer value to be preceding the time units (i.e. "1minute"). However, this does not align with the function's definition, or the [Graphite web implementation](https://github.com/graphite-project/graphite-web/blob/b52987ac97f49dcfb401a21d4b92860cfcbcf074/webapp/graphite/render/functions.py#L5258) (see [here](https://github.com/graphite-project/graphite-web/blob/d6bda76332381a2ff8487efb46f386ad7c6e9bda/webapp/graphite/render/attime.py#L192) for the time parsing function). In order to align with Graphite web's behavior, smartSummarize has been modified to prepend a value of "1" before the interval string, in order to use parser.GetIntervalString without an error being returned because there is no preceding integer.

### DIFF
--- a/expr/functions/smartSummarize/function.go
+++ b/expr/functions/smartSummarize/function.go
@@ -70,7 +70,8 @@ func (f *smartSummarize) Do(ctx context.Context, e parser.Expr, from, until int6
 	start := args[0].StartTime
 	stop := args[0].StopTime
 	if alignToInterval != "" {
-		interval, err := parser.IntervalString(alignToInterval, 1)
+		alignTo := "1" + alignToInterval // Add a 1 before the alignTo interval, so that IntervalString properly parses it
+		interval, err := parser.IntervalString(alignTo, 1)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


carrieedwards:
@Civil FYI, I made this change to match Graphite web's behavior, but I am a little bit concerned that if users expect to be able to pass in an interval such as "1year", then this change will cause unexpected interval differences. Let me know what you think.

npazosmendez (dismissed):

npazosmendez (dismissed):

npazosmendez (approved):

PR #733, opened 2022-11-02
Fix consolidateFunc setting and add test for consolidateBy

carrieedwards:
This PR adds tests for the consolidateBy function, as there previously were no tests. It also updates the Name field of the results to match how Graphite web's consolidateBy function assigns Name to the results (see [Graphite web's implementation](https://github.com/graphite-project/graphite-web/blob/b52987ac97f49dcfb401a21d4b92860cfcbcf074/webapp/graphite/render/functions.py#L2101)).

Additionally, it appeared that several functions, such as constantLine, time, and identity, had a consolidationFunc setting that was not aligned with Graphite web, so this was fixed.

### DIFF
--- a/expr/functions/consolidateBy/function_test.go
+++ b/expr/functions/consolidateBy/function_test.go
@@ -1,0 +1,83 @@
+package consolidateBy
+
+import (
+	"github.com/go-graphite/carbonapi/expr/helper"
+	"github.com/go-graphite/carbonapi/expr/metadata"
+	"github.com/go-graphite/carbonapi/expr/types"
+	"github.com/go-graphite/carbonapi/pkg/parser"
+	th "github.com/go-graphite/carbonapi/tests"
+	"testing"
+	"time"
+)
+
+func init() {
+	md := New("")
+	evaluator := th.EvaluatorFromFunc(md[0].F)
+	metadata.SetEvaluator(evaluator)
+	helper.SetEvaluator(evaluator)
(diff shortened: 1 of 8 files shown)


expr/functions/consolidateBy/function_test.go
```diff
@@ -0,0 +1,83 @@
...
+	now32 := time.Now().Unix()
+
+	tests := []th.EvalTestItem{
+		{
+			"consolidateBy(metric1,\"sum\")",
+			map[parser.MetricRequest][]*types.MetricData{
+				{"metric1", 0, 1}: {types.MakeMetricData("metric1", []float64{1, 2, 3, 4, 5}, 1, now32)},
+			},
+			[]*types.MetricData{types.MakeMetricData("consolidateBy(metric1,\"sum\")",
+				[]float64{1, 2, 3, 4, 5}, 1, now32).SetTag("consolidateBy", "sum").SetConsolidationFunc("sum")},
```
npazosmendez, line 33: I think we would need to add the `actual.ConsolidateFunc != want.ConsolidateFunc` check in this helper for the assertion to be made

https://github.com/go-graphite/carbonapi/blob/50b983a0ac3d230c88067b151b553c92a2be41ef/tests/helper.go#L399

npazosmendez (dismissed):

npazosmendez (approved):
LGTM

PR #741, opened 2022-11-21
response/backend cache short duration tune

msaf1980:
- Change long cache duration to 3 hours and 2 minute until offset by default, also make until offset tunable.
- Recalc duration on from/until timestamps truncate.


Civil:
It would be great to have a few lines somewhere in the docs about what is ShortTimeout and ShortDuration, as names of the options are not self-explanatory.

Civil (approved):

msaf1980:
Documentation updated.

Civil (approved):

PR #742, opened 2022-11-26
Refactor HTTP client

Civil:
* Add preliminary support for mTLS (all required TLS options)
 * Remove CarbonSearch support
 * Fix Idle Connection Timeout for HTTP Client (shouldn't reuse read timeout as it was before)
 * Move all common part for protocols to helper pkg


zipper/types/tlsconfig.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+type ClientCertificatePairs struct {
+	CertFile       string `mapstructure:"certFile"`
+	PrivateKeyFile string `mapstructure:"privateKeyFile"`
+}
+
+type TLSConfig struct {
+	CACertFiles      []string                 `mapstructure:"caCertFiles"`
+	CertificateParis []ClientCertificatePairs `mapstructure:"certificateParis"`
```
msaf1980, line 16: CertificateParis -> CertificatePairs ?

msaf1980 (approved):

PR #750, opened 2023-01-30
Fix strftime

jhesketh:
strftime has moved to github (as far as I can tell).

Also bump its version while I'm here.


deniszh:
Hmmm. mTLS tests are failing. Not sure is it related to change.

Civil:
Its not related, I think I initially thought to create test certificate that would be valid for 10+ years, but did a mistake and its expired in January, so mTLS tests with strict checking now fails.

Civil (approved):

PR #759, opened 2023-04-12
Fix hitcount function handling of intervals and alignToInterval 

carrieedwards:
This PR is a rewrite of the hitcount function in order to match Graphite web's behavior. There are two major issues that were discovered: 

1) When the `interval` parameter was set to a value smaller than the fetched data's step between data points, it results were incorrectly shifted left, such that all of the hit counts were concentrated towards the left buckets, and there were empty buckets on the right. For example:

hitcount(metric1, "6m")
[]*types.MetricData{{"metric1", 0, 1}: {types.MakeMetricData("metric1", []float64{2, 4, 6}, 600, 100)},} // 10m step

The result will be []float64{1200, 2400, 3600, 0, 0, 0}. The hit values are all concentrated on the left half of the buckets.

2) Upon looking at Graphite web's implementation of hitcount, when the `alignToInterval` parameter is set to true, the start time is adjusted and the data is re-fetched. In CarbonAPI, the start time was adjusted, but data was not re-fetched. 

**Major changes in this PR:**
- The hitcount function was completely rewritten to match the behavior of Graphite web. This helps address the issue with intervals smaller than the data's step, as well as fixes discrepancies in results that were occurring with CarbonAPI tests that were tested in Graphite web. 
- The Metrics() function in pkg/parser.go was updated to pass in the from/to of the request, and code was added to determine the correct start time for the data being fetched for the hitcount function when `alignToInterval ` is set to true. The new start time is determined based on the request's from value and the interval specified. This is intended to prevent needing to re-fetch the data as is done in Graphite web.

### DIFF
--- a/expr/functions/hitcount/function.go
+++ b/expr/functions/hitcount/function.go
@@ -3,16 +3,14 @@ package hitcount
 import (
 	"context"
 	"fmt"
-	"math"
-	"strconv"
-	"strings"
-	"time"
-
 	"github.com/go-graphite/carbonapi/expr/helper"
 	"github.com/go-graphite/carbonapi/expr/interfaces"
 	"github.com/go-graphite/carbonapi/expr/types"
 	"github.com/go-graphite/carbonapi/pkg/parser"
 	pb "github.com/go-graphite/protocol/carbonapi_v3_pb"
+	"math"
+	"strconv"
+	"strings"
 )
 
 type hitcount struct {
@@ -105,8 +103,6 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
 		buckets := make([][]float64, bucketCount)
 		newStart := stop - bucketCount*interval
 		r.StartTime = newStart
-		newStartDateTime := time.Unix(newStart, 0)
-		fmt.Println("New start time is: ", newStartDateTime)
 
 		for i, v := range arg.Values {
 			if math.IsNaN(v) {
--- a/expr/functions/hitcount/function.go
+++ b/expr/functions/hitcount/function.go
@@ -52,7 +52,7 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
 	if err != nil {
 		return nil, err
 	}
-	bucketSize := int64(bucketSizeInt32)
+	interval := int64(bucketSizeInt32)
 
 	alignToInterval, err := e.GetBoolNamedOrPosArgDefault("alignToInterval", 2, false)
 	if err != nil {
@@ -62,10 +62,9 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
 	start := args[0].StartTime
 	stop := args[0].StopTime
 	if alignToInterval {
-		start = helper.AlignStartToInterval(start, stop, bucketSize)
+		start = helper.AlignStartToInterval(start, stop, interval)
 	}
 
-	buckets := helper.GetBuckets(start, stop, bucketSize)
 	results := make([]*types.MetricData, 0, len(args))
 	for _, arg := range args {
 		var nameBuf strings.Builder
@@ -82,11 +81,13 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
 		}
 		nameBuf.WriteString(")")
 
+		bucketCount := helper.GetBuckets(start, stop, interval)
+
 		r := &types.MetricData{
 			FetchResponse: pb.FetchResponse{
 				Name:              nameBuf.String(),
-				Values:            make([]float64, buckets, buckets+1),
-				StepTime:          bucketSize,
+				Values:            make([]float64, bucketCount, bucketCount+1),
+				StepTime:          interval,
 				StartTime:         start,
 				StopTime:          stop,
 				ConsolidationFunc: "max",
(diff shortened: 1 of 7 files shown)


expr/functions/hitcount/function.go
```diff
@@ -95,41 +101,60 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
...
-			if !math.IsNaN(v) {
-				if math.IsNaN(count) {
-					count = 0
-				}
+		step := arg.StepTime
+		buckets := make([][]float64, bucketCount)
+		newStart := stop - bucketCount*interval
+		r.StartTime = newStart
+		newStartDateTime := time.Unix(newStart, 0)
+		fmt.Println("New start time is: ", newStartDateTime)
```
leizor, line 109: Is this a debugging print statement?

pkg/parser/parser.go
```diff
@@ -212,6 +212,37 @@ func (e *expr) Metrics() []MetricRequest {
...
+				if alignToInterval {
+					bucketSizeInt32, err := e.GetIntervalArg(1, 1)
+					if err != nil {
+						return nil
+					}
+					interval := int64(bucketSizeInt32)
+					// This is done in order to replicate the behavior in Graphite web when alignToInterval is set,
+					// in which new data is fetched with the adjusted start time.
+					for i, _ := range r {
+						start := from
```
npazosmendez, line 234: I think this needs to take the `From` from the current request; otherwise the offsets from nested functions would be overwritten.  Similar to what is done with the moving functions: https://github.com/go-graphite/carbonapi/pull/759/files#diff-2a6035c752559ce0fb74ced451e9e4363c01598c669d8f77a46be9450e6f6d87L211
```suggestion
						start := r[i].From
```

pkg/parser/parser_test.go
```diff
@@ -490,3 +490,64 @@ func TestDoGetIntArg(t *testing.T) {
...
+			1410346740,
+			1410346865,
+			[]MetricRequest{
+				{
+					Metric: "metric1",
+					From:   1410346740,
+					Until:  1410346865,
+				},
+			},
+		},
```
npazosmendez, line 544: Here is a test that should catch the problem I mentioned in the other comment:

```go
		{
			"hitcount(timeShift(metric1, '-1h'),'1h')",
			&expr{
				target: "hitcount",
				etype:  EtFunc,
				args: []*expr{
					{
						target: "timeShift",
						etype:  EtFunc,
						args: []*expr{
							{target: "metric1"},
							{valStr: "-1h", etype: EtString},
						},
						argString: "metric1, '-1h'",
					},
					{valStr: "1h", etype: EtString},
					{valStr: "true", etype: EtBool},
				},
				argString: "timeShift(metric1, '-1h'),'1h'",
			},
			1410346740,
			1410346865,
			[]MetricRequest{
				{
					Metric: "metric1",
					From:   1410339600,
					Until:  1410343265,
				},
			},
		},
```

expr/functions/hitcount/function_test.go
```diff
@@ -55,11 +55,12 @@ func TestHitcount(t *testing.T) {
...
 			},
-			[]float64{35, 70, 105, 140, math.NaN(), 25},
+			[]float64{5, 40, 75, 110, 120, 25},
+
 			"hitcount(metric1,'30s')",
 			30,
-			now32,
-			now32 + 31*5,
+			1410344975,
+			1410345155,
```
npazosmendez, line 63: These should hold that end = from +count*step
```suggestion
			1410344975 + 31*5,
```

expr/functions/hitcount/function_test.go
```diff
@@ -72,8 +73,8 @@ func TestHitcount(t *testing.T) {
 			[]float64{375},
 			"hitcount(metric1,'1h')",
 			3600,
-			tenFiftyNine,
-			tenFiftyNine + 25*5,
+			1410343265,
+			1410346865,
```
npazosmendez, line 77:
```suggestion
			1410343265 + 25*5,
```

expr/functions/hitcount/function_test.go
```diff
@@ -83,11 +84,11 @@ func TestHitcount(t *testing.T) {
...
 					5}, 5, tenFiftyNine)},
 			},
-			[]float64{105, 270},
+			[]float64{375},
 			"hitcount(metric1,'1h',true)",
 			3600,
-			tenFiftyNine - (59 * 60),
-			tenFiftyNine + 25*5,
+			1410346740,
+			1410350340,
```
npazosmendez, line 91:
```suggestion
			1410346740 + 25*5,
```

expr/functions/hitcount/function_test.go
```diff
@@ -97,11 +98,24 @@ func TestHitcount(t *testing.T) {
...
```
npazosmendez, line 105:
```suggestion
			1410346740 + 25*5,
```

expr/functions/hitcount/function.go
```diff
@@ -95,41 +99,58 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
...
+				}
 			}
 		}
-
-		// remaining values
-		if bucketItems > 0 {
-			r.Values[ridx] = count
+		newValues := make([]float64, bucketCount)
+		for i, bucket := range buckets {
+			if bucket != nil {
```
npazosmendez, line 143: `bucket` should never be `nil` AFAICT.  I think what [graphite web is doing](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L5509) is checking if the bucket is empty, so:

```suggestion
			if len(bucket) == 0 {
```

npazosmendez (commented):
Some comments, but looks good overall

expr/functions/hitcount/function.go
```diff
@@ -82,11 +84,13 @@ func (f *hitcount) Do(ctx context.Context, e parser.Expr, from, until int64, val
...
 		r := &types.MetricData{
 			FetchResponse: pb.FetchResponse{
 				Name:              nameBuf.String(),
-				Values:            make([]float64, buckets, buckets+1),
-				StepTime:          bucketSize,
+				Values:            make([]float64, bucketCount, bucketCount+1),
+				StepTime:          interval,
 				StartTime:         start,
 				StopTime:          stop,
 				ConsolidationFunc: "max",
```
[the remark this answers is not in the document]
npazosmendez: No, to the `ConsolidationFunc`
```go
				ConsolidationFunc: "max",
```

ywwg (dismissed):
nitting

npazosmendez (dismissed):
LGTM

npazosmendez (approved):
still LGTM

Civil (approved):

PR #789, opened 2023-07-19
fix return code for render_handler

Thorsieger:
Hello,

carbonapi do not response correct code 404 when no metric is found from gocarbon/carbonserver when requesting for a render

This should fix the issue : 

returnCode is reasigned in an other scope, so any modification won't have effect ouside the if


msaf1980 (approved):

msaf1980:
> Hello,
> 
> carbonapi do not response correct code 404 when no metric is found from gocarbon/carbonserver when requesting for a render
> 
> This should fix the issue :
> 
> returnCode is reasigned in an other scope, so any modification won't have effect ouside the if

May be write e2e test for check error ?

PR #793, opened 2023-08-23
feat(functions): add support for consolidateBy

lordvidex:
consolidateBy should send argument to backend so that metrics can be rollup using the aggregation function specified in the function.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -38,9 +38,16 @@ func (eval evaluator) Fetch(ctx context.Context, exprs []parser.Expr, from, unti
 				MaxDataPoints:  maxDataPoints,
 			}
 			metricRequest := parser.MetricRequest{
-				Metric: fetchRequest.PathExpression,
-				From:   fetchRequest.StartTime,
-				Until:  fetchRequest.StopTime,
+				Metric:            fetchRequest.PathExpression,
+				From:              fetchRequest.StartTime,
+				Until:             fetchRequest.StopTime,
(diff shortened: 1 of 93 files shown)


lordvidex:
The first commit contains changes that should be reviewed while the second are refactor changes to make code compile properly.

npazosmendez:
Hello!

> consolidateBy should send argument to backend so that metrics can be rollup using the aggregation function specified in the function.

Are you suggesting to add this as an extension? AFAIK stock Graphite Web (+Whisper) does not support this. The aggregation function is already set and fixed on write time (through `storage-aggregation.conf`). The consolidation function is only used for post-rendering aggregation, to obey the max data points argument.

deniszh:
Graphite web have support for transferring target to backend, and it was
used in some non standard backends. Just FYI.

ср, 23 авг. 2023 г., 16:56 Nicolás Pazos ***@***.***>:

npazosmendez:
Right, thanks for clarifying @deniszh . So, to be clear, this is indeed a non standard extension, right? 

BTW I don't think it's bad, but it's good to keep track of how carbonapi diverges from stock graphite, and put feature flags when desirable.

Civil (approved):

npazosmendez:
Probably closable since https://github.com/go-graphite/carbonapi/pull/830 was merged

Civil:
Yeah, I'll close this one. But if you merged PR doesn't do what you think it should do - please create a new one.

PR #811, opened 2023-12-18
expr: clean from direct use zipper/limiter from carbonapi app global config

msaf1980:
Try to refactor expr package for avoid direct usage zipper/limiter from carbonapi app global config.
Not a complete refactor, no separate Eval/Fetch (may be in future).


expr/expr.go
```diff
@@ -6,22 +6,31 @@ import (
...
-type evaluator struct{}
+type evaluator struct {
+	limiter limiter.SimpleLimiter
+	zipper  zipper.CarbonZipper
+}
 
 func (eval evaluator) Fetch(ctx context.Context, exprs []parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) (map[parser.MetricRequest][]*types.MetricData, error) {
-	if err := config.Config.Limiter.Enter(ctx); err != nil {
+	if eval.zipper == nil {
+		// TODO: may be return error or not check. But for use expr in external applications without CarbonZipper implementation
```
msaf1980, line 26: For external applications may be better to implement CarbonZipper.

msaf1980:
Reason for do this - use expr methods in separate product. Use for configure it with carbonapi app config is bad design.

msaf1980:
@npazosmendez  Another variant of expr refactor. It's compromise (global evaluator exist, but only in expr). May be less clear, but more simple and no Do changes is required.

msaf1980:
Not complete solution
> There's other cases that are even more tricky, or maybe even impossible. For example [aliasQuery](https://github.com/go-graphite/carbonapi/blob/main/expr/functions/aliasQuery/function.go#L0-L1)

PR #814, opened 2023-12-28
fix(render): return error on partial targets fetch

msaf1980:


cmd/carbonapi/http/render_handler.go
```diff
@@ -327,6 +323,9 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
 				result, err := expr.FetchAndEvalExp(ctx, config.Config.Evaluator, exp, from32, until32, values)
 				if err != nil {
 					errors[target] = merry.Wrap(err)
+					// if config.Config.Upstreams.RequireSuccessAll {
```
almostinf, line 326: Did you forget to put it away or is that what it's supposed to be?

cmd/carbonapi/http/render_handler.go
```diff
@@ -347,19 +346,19 @@ func renderHandler(w http.ResponseWriter, r *http.Request) {
...
 		returnCode, errMsgs = helper.MergeHttpErrorMap(errors)
 		logger.Debug("error response or no response", zap.Strings("error", errMsgs))
 		// Allow override status code for 404-not-found replies.
-		if returnCode == 404 {
+		if returnCode == http.StatusNotFound {
 			returnCode = config.Config.NotFoundStatusCode
 		}
 
-		if returnCode == 400 || returnCode == http.StatusForbidden || returnCode >= 500 {
+		if returnCode == http.StatusBadRequest || returnCode == http.StatusNotFound || returnCode == http.StatusForbidden || returnCode >= 500 {
```
almostinf, line 361: Can we also change 500 to `http.StatusInternalServerError'?

msaf1980: Check againt range with mnemonic constant is bad idea, in my opinion.

mchrome (approved):

Civil (approved):

PR #818, opened 2024-01-24
expr: refactor for avoid global evaluator usage

msaf1980:
Try to refactor expr package for avoid global evaluator usage


expr/expr.go
```diff
@@ -128,63 +136,19 @@ func (eval evaluator) Eval(ctx context.Context, exp parser.Expr, from, until int
...
+		return nil, ErrZipperNotInit
 	}
-
-	return res, errors
+	return &Evaluator{limiter: limiter, zipper: zipper}, nil
 }
 
-// EvalExpr is the main expression evaluator.
-func EvalExpr(ctx context.Context, e parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types.MetricData, error) {
+// EvalExpr is the main expression Evaluator.
```
Tetrergeru, line 150: I think the comment got renamed by accident

```suggestion
// EvalExpr is the main expression evaluator.
```

expr/functions/aggregate/function_test.go
```diff
@@ -7,19 +7,20 @@ import (
...
+var (
+	md []interfaces.FunctionMetadata = New("")
+)
+
 func init() {
-	md := New("")
-	evaluator := th.EvaluatorFromFunc(md[0].F)
-	metadata.SetEvaluator(evaluator)
-	helper.SetEvaluator(evaluator)
+	md = New("")
```
Tetrergeru, line 23: Do we need this? This variable is initialized the same way in declaration block

npazosmendez:
Hey @msaf1980, thanks for the ping. I will take a look at this later today.

npazosmendez:
This looks reasonable. I agree that the big refactor of `Do`s is not ideal, but I can't think of any better options.

One question: did the global evaluator bring you any particular problems? I agree it's not a nice pattern, but I wonder what motivated you in such big refactor.

msaf1980:
> o
Changing signature of Do - is ugly, i think.
Compromise variant is https://github.com/go-graphite/carbonapi/pull/811
No Do change, but helper cleaned from global vars (duplicate in two packages expr and helper)
And may be better.

Reason for do this - use expr methods in separate product. Use for configure it with carbonapi app config is bad design.
May be in future we separate Fetch and Eval (no refetch on it), but for moving function it's not too easy (we need precalc points step).

npazosmendez:
> May be in future we separate Fetch and Eval (no refetch on it), but for moving function it's not too easy (we need precalc points step).

There's other cases that are even more tricky, or maybe even impossible. For example [aliasQuery](https://github.com/go-graphite/carbonapi/blob/main/expr/functions/aliasQuery/function.go#L0-L1)

npazosmendez (approved):
LGTM, but another review would be nice since this is a big PR

msaf1980:
> > May be in future we separate Fetch and Eval (no refetch on it), but for moving function it's not too easy (we need precalc points step).
> 
> There's other cases that are even more tricky, or maybe even impossible. For example [aliasQuery](https://github.com/go-graphite/carbonapi/blob/main/expr/functions/aliasQuery/function.go#L0-L1)

We planned separate in some steps:
1) parse
2) extract side effects (get step and correct from/until (for moving)/aliasQuery need fetches/etc
3) Fetch
4) Evaluate

But it's not easy and require a lot of time - get step require changes in Storage API.

May be change Do and pass-in evaluator is a simple solution at now with hard refactor in future.

Civil (approved):

PR #824, opened 2024-02-19
fix(main): gracefully shutdown http servers

mchrome:


cmd/carbonapi/main.go
```diff
@@ -151,6 +154,17 @@ func main() {
 
 				wg.Done()
 			}(listener, isTLS)
+
+			go func() {
+				stop := make(chan os.Signal, 1)
+				signal.Notify(stop, syscall.SIGTERM, syscall.SIGINT)
```
msaf1980, line 160: Don't set signal handler in a loop. More better - set one handler and shutdown all listeners in them.

msaf1980 (approved):

Issue #791, opened 2023-08-10
[BUG] please revert 73b1eb298bdcba488391b087ebb91cd24560cd45 it fully broke tags query 


tantra35:
After upgrade to last main we discover that in grafana for variables tags queries fully broken, we doesn't see any tags as expected, after some investigation we found that the reason is 73b1eb298bdcba488391b087ebb91cd24560cd45, so please revert it

Civil:
Done, but it would be great if you'll be able to provide more information (log, error messages, etc).

And I hope @msaf1980 would be able to have a look at it as that was his PR.

msaf1980:
@tantra35 Please past failed queries. I don't know why it's may be failed, but add some e2e tests for check.

PR #826, opened 2024-03-14
tags/autoComplete: return detailed error code instead of 500

msaf1980:


msaf1980:
Reincarnatin of https://github.com/go-graphite/carbonapi/commit/73b1eb298bdcba488391b087ebb91cd24560cd45
But reveted with strange error https://github.com/go-graphite/carbonapi/issues/791
No problem description from user. For verify results e2e tests added.

Civil (approved):

PR #830, opened 2024-04-11
feat(functions): add a config option to pass consolidateBy to the storage backend

mchrome:
This PR is mostly a copy of #793.
Passes consolidateBy with its argument to the backend through the FilteringFunctions field. Only works with carbonapi_v3_pb protocol.
Fixes broken tests and removes passing ConsolidationFunc field further after parsing the target in Fetch. This is needed since this field is empty during evaluation and "values" map will not return anything.
In addition I added an option to the config to enable this behavior.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -48,9 +48,16 @@ func (eval Evaluator) Fetch(ctx context.Context, exprs []parser.Expr, from, unti
 				MaxDataPoints:  maxDataPoints,
 			}
 			metricRequest := parser.MetricRequest{
-				Metric: fetchRequest.PathExpression,
-				From:   fetchRequest.StartTime,
-				Until:  fetchRequest.StopTime,
+				Metric:            fetchRequest.PathExpression,
+				From:              fetchRequest.StartTime,
+				Until:             fetchRequest.StopTime,
+				ConsolidationFunc: m.ConsolidationFunc,
+			}
+
+			if m.ConsolidationFunc != "" {
+				fetchRequest.FilterFunctions = append(fetchRequest.FilterFunctions, &pb.FilteringFunction{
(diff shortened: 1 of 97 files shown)


expr/expr.go
```diff
@@ -53,6 +54,13 @@ func (eval Evaluator) Fetch(ctx context.Context, exprs []parser.Expr, from, unti
 				Until:  fetchRequest.StopTime,
 			}
 
+			if eval.passFunctionsToBackend && m.ConsolidationFunc != "" {
+				fetchRequest.FilterFunctions = append(fetchRequest.FilterFunctions, &pb.FilteringFunction{
+					Name:      "consolidateBy",
```
almostinf, line 59: здесь также предлагаю "consolidateBy" вынести в константу

Civil (approved):

expr/expr.go
Civil: @almostinf nit: 
многие контрибьюторы в carbonapi не знают <подставить сюда название человеческого языка>, поэтому я предлагаю использовать английский в качестве основного для коммуникаций в комментариях, issues и код-ревью, естественно за исключением ситуации когда вы не знаете английского.

English:
there are a lot of contributors to carbonapi don't know ${language_name}, therefore I'd strongly suggest everyone to use English as a common language for comments and issue descriptions and PR reviews, unless you don't know it.


@mchrome about that comment - I think it doesn't matter much if you'll move out "consolidateBy" to const or not. When it is a single function that can be passed through - it is ok to use whatever. When we'll have more - we'll need to come up with some easy to use way to manage that anyway and probably would need to refactor that code again, no matter what.

mchrome: I think I'll keep it like that then.

PR #831, opened 2024-04-25
Update all vendored dependencies

Civil:
One small change was required - for irondb, Discover config option was removed at some point. Commented it out.


npazosmendez (approved):

Hipska:
May I ask why the modules are also included in the `vendor` dir? This isn't a usual practice with most other go applications.

Civil:
Historical reasons. That project predates modules and during early days having all dependencies in the repo was the mainstream way to go. It have some small benefits - e.x. it guarantees that build will be reproducible and original dev won't disable that with force-push or something.

PR #833, opened 2024-05-20
PromethizeTagValue panic on empty value fix

artem1996:


zipper/protocols/prometheus/helpers/helpers.go
```diff
@@ -139,29 +139,42 @@ func AdjustStep(start, stop, maxPointsPerQuery, minStep int64, forceMinStepInter
...
 func PromethizeTagValue(tagValue string) (string, types.Tag) {
 	// Handle = and =~
-	t := types.Tag{}
-	idx := strings.Index(tagValue, "=")
-	if idx != -1 {
-		if tagValue[idx+1] == '~' {
+	var (
+		t       types.Tag
+		tagName string
+		idx     = strings.Index(tagValue, "=")
```
Civil, line 145: I'm not a big fan of using val() block to initialize valuies, but I don't think it is a major problem of the PR. More of a FIY for next time.

Civil (approved):

PR #834, opened 2024-06-13
fix: validate consolidateBy arguments

mchrome:
Add validation for consolidateBy arguments when `passFunctionsToBackend: true`. [Valid aggregation functions.](https://graphite.readthedocs.io/en/stable/functions.html#graphite.render.functions.consolidateBy)

### DIFF
--- a/expr/functions/consolidateBy/function.go
+++ b/expr/functions/consolidateBy/function.go
@@ -26,6 +26,17 @@ func New(configFile string) []interfaces.FunctionMetadata {
 	return res
 }
 
+// The set of valid aggregation methods for consolidateBy function.
+var ValidAggregateFunctions = map[string]struct{}{
+	"average": {},
+	"avg":     {},
+	"max":     {},
+	"min":     {},
+	"sum":     {},
+	"first":   {},
+	"last":    {},
+}
+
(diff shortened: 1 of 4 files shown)


expr/functions/consolidateBy/function.go
```diff
@@ -26,6 +26,16 @@ func New(configFile string) []interfaces.FunctionMetadata {
 	return res
 }
 
+var ValidAggregateFunctions = map[string]struct{}{
```
kissken, line 29: godoc?

expr/functions/consolidateBy/function.go
```diff
@@ -26,6 +26,16 @@ func New(configFile string) []interfaces.FunctionMetadata {
...
+	"average": {},
+	"avg":     {},
+	"max":     {},
+	"min":     {},
+	"sum":     {},
+	"first":   {},
+	"last":    {},
+}
```
kissken, line 37: "mean" should not be?

mchrome: According to the [documentation](https://graphite.readthedocs.io/en/stable/functions.html#graphite.render.functions.consolidateBy) these are all of them.

msaf1980 (approved):

PR #842, opened 2024-09-06
MaxDataPoints consolidation improvements: support nudging for consistent bucketing

npazosmendez:
Brings in these features: https://github.com/grafana/carbonapi/pull/146

________________________

## NudgeStartTimeOnAggregation

I've described the purpose of this in the code. Here's a copy of the relevant parts:

```
// NudgeStartTimeOnAggregation enables nudging the start time of metrics
// when aggregated. The start time is nudged in such way that timestamps
// always fall in the same bucket. This is done by GraphiteWeb, and is
// useful to avoid jitter in graphs when refreshing the page.

[ ... ]

// nudgePointsCount returns the number of points to discard at the beginning of
// the series when aggregating. This is done if NudgeStartTimeOnAggregation is
// enabled, and has the purpose of assigning timestamps of a series to buckets
// consistently across different time ranges. To simplifiy the aggregation
// logic, we discard points at the beginning of the series so that a bucket
// starts right at the beginning. This function calculates how many points to
// discard.

```

 TLDR: if the consolidation is done naively (as it is being done now) , the same data point can be assigned to different buckets depending on the start time of the series. This can cause jitter in graphs when refreshing constantly.

This is done by graphite web here: https://github.com/graphite-project/graphite-web/blob/dca59dc72ae28ffdae659232c10c60aa598536eb/webapp/graphite/render/views.py#L225

and by Metrictank here: https://github.com/grafana/metrictank/blob/309ba74749c4fd8c60950929ff0719af8abc0799/consolidation/consolidate.go#L26

My implementation and tests are mainly inspired on Metrictank.

## UseBucketsHighestTimestampOnAggregation

Additionally, `UseBucketsHighestTimestampOnAggregation` let's the user configure the bucket's timestimes in such way that the upper bound is used instead of the lower bound. This was also done by Metrictank, and its purpose is to avoid results from appearing to predict the future (i.e. a spike in "real time" ts=6 could be moved to a ts=4). 

Both of these features are disabled by default for now to avoid unexpected changes on upgrades.

_______________________________

Some videos showing this:

#### CarbonAPI with nudging disabled

https://github.com/grafana/carbonapi/assets/32206519/1d285484-3e4b-4551-bed4-cd82908241ab

Notice the sudden change when the values change buckets

#### CarbonAPI with nudging enabled

https://github.com/grafana/carbonapi/assets/32206519/4d57699b-318b-41c0-a367-0643d7496d4c

#### Metrictank's version (which was my ground truth here)

https://github.com/grafana/carbonapi/assets/32206519/860cd2f6-85b2-42e8-ab8e-07cb30a7e121

### DIFF
--- a/cmd/carbonapi/config/config.go
+++ b/cmd/carbonapi/config/config.go
@@ -113,6 +113,9 @@ type ConfigType struct {
 	MaxQueryLength              uint64 `mapstructure:"maxQueryLength"`
 	CombineMultipleTargetsInOne bool   `mapstructure:"combineMultipleTargetsInOne"`
 
+	NudgeStartTimeOnAggregation             bool `mapstructure:"nudgeStartTimeOnAggregation"`
+	UseBucketsHighestTimestampOnAggregation bool `mapstructure:"useBucketsHighestTimestampOnAggregation"`
+
 	ResponseCache cache.BytesCache `mapstructure:"-" json:"-"`
 	BackendCache  cache.BytesCache `mapstructure:"-" json:"-"`
 
--- a/cmd/carbonapi/config/init.go
+++ b/cmd/carbonapi/config/init.go
@@ -26,6 +26,7 @@ import (
 	fconfig "github.com/go-graphite/carbonapi/expr/functions/config"
 	"github.com/go-graphite/carbonapi/expr/helper"
 	"github.com/go-graphite/carbonapi/expr/rewrite"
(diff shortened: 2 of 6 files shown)


Civil (approved):

Civil:
I'll flip those two flags to "true" by default in master, as overall idea is to be as compatible with graphite-web as feasible out of the box.

PR #843, opened 2024-10-04
doc: changelog for version 0.17.0

Tetrergeru:
Add changelog and update compatibility matric for the new version

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,29 @@ Changes
 
 CHANGELOG
 ---------
+
+**0.17.0**
+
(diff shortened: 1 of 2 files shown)


Civil (approved):

Hipska:
@Tetrergeru @Civil 0.17.0 also doesn't seem to be marked as latest version?

Tetrergeru:
@Hipska, I decided not to mark release as latest before review. I can do it now, if you approve

Hipska:
Now it is too late, since there is already an 0.17.0.1 marked as latest 😉

PR #848, opened 2024-11-29
fix SummarizeValues to return last non-NaN value

spacefreak86:
When `SummarizeValues` is called with `last` or `current` as the aggregation function, it returns the last value of the series, whether it is NaN or not. This causes `legendValue` to output "NaN" in the legend, even if there are non-NaN values the series.

This PR changes the behavior of `SummarizeValues` so that the last non-NaN value is returned.

### DIFF
--- a/expr/consolidations/consolidations.go
+++ b/expr/consolidations/consolidations.go
@@ -163,7 +163,7 @@ func SummarizeValues(f string, values []float64, XFilesFactor float32) float64 {
 				total++
 			}
 		}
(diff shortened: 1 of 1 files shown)


spacefreak86:
Could anybody take a look at this please? It is not the fault of this PR that the DeepSource test fails, it does not change the complexity of the SummarizeValues function.

Civil (approved):

spacefreak86:
@Civil could you please review again once more?

PR #928, opened 2026-05-18
Check interval arg is within int32 bounds

fionaliao:
Currently `IntervalString` is parsed as an `int` and then casted to an `int32`, which could result in an overflow. This was spotted when a `summarize()` query was run with a `100y` interval - in this case, the cast caused a negative value to be returned, and then caused a panic within `summarize.Do` as it caused a negative index to be accessed in a slice.

Proposing to fix by erroring if the parsed value is out of the int32 bounds instead of silently returning a possibly incorrect value.

### DIFF
--- a/expr/functions/summarize/function_test.go
+++ b/expr/functions/summarize/function_test.go
@@ -2,6 +2,7 @@ package summarize
 
 import (
 	"context"
+	"errors"
 	"math"
 	"testing"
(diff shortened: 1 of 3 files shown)


carrieedwards (approved):
LGTM

deniszh (approved):

PR #930, opened 2026-05-26
Ignore boolean value for alignTo parameter in smartSummarize

carrieedwards:
This PR updates the smartSummarize function, allowing it to ignore the alignTo parameter if a boolean value is passed in. This is for compatibility with graphite-web. For context, smartSummarize used to accept a boolean alignToFrom parameter, which has since been deprecated. However, graphite-web does not error out if a boolean is passed in to the smartSummarize function for the last argument/alignTo value. 

See: https://github.com/graphite-project/graphite-web/blob/b5d272be6abdf358aebe56d12df777c2c97de4c3/webapp/graphite/render/functions.py#L5280

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -141,6 +141,20 @@ func (e *expr) NamedArg(name string) (Expr, bool) {
 	return expr, exist
 }
 
+// NamedOrPosArg returns the named argument when present, otherwise the positional argument.
+func NamedOrPosArg(e Expr, name string, pos int) (Expr, bool) {
+	if e == nil || e.IsInterfaceNil() {
+		return nil, false
+	}
+	if arg, ok := e.NamedArg(name); ok {
+		return arg, true
+	}
+	if e.ArgsLen() > pos {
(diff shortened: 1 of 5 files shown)


pkg/parser/parser.go
```diff
@@ -141,6 +141,20 @@ func (e *expr) NamedArg(name string) (Expr, bool) {
 	return expr, exist
 }
 
+// NamedOrPosArg returns the named argument when present, otherwise the positional argument.
+func NamedOrPosArg(e Expr, name string, pos int) (Expr, bool) {
+	if e == nil || e.IsInterfaceNil() {
+		return nil, false
+	}
```
npazosmendez: Two comments:
1) Could we make this a method on `Expr` to match `(e *expr) NamedArgs() ` and other existing ones?
2) Any reason to check for null here? I don't see that being done anywhere else, I don't think we expect that to be true

npazosmendez (commented):
Looks good, just a minor comment

npazosmendez (approved):

PR #931, opened 2026-05-27
fix: support more time formats and use in `timeSlice`

npazosmendez:
Two changes:

- Expand time parsing (`ParseAtTime`) with more cases that stock graphite supports:
	- Support expressions like `today-2d`
	- Case-insensitiveness
	- Support 4-digit years like `12/11/1991`
- Make `timeSlice` use such parsing. Currently it's treating the arguments as intervals, which are not (e.g. `3m` is read as 3 minutes after epoch). For retro compatibility, the previous logic is kept as a fallback when the parsing fails.  

For reference, see https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/attime.py

### DIFF
--- a/date/date.go
+++ b/date/date.go
@@ -12,11 +12,14 @@ import (
 var errBadTime = errors.New("bad time")
 var timeNow = time.Now
 
-// MockTimeNow replaces the package clock and returns the previous one so tests can restore it.
-func MockTimeNow(f func() time.Time) func() time.Time {
+// MockTimeNow replaces the package clock for tests and returns a restore
+// function meant to be deferred:
+//
+//	defer date.MockTimeNow(func() time.Time { ... })()
+func MockTimeNow(f func() time.Time) func() {
 	prev := timeNow
 	timeNow = f
-	return prev
+	return func() { timeNow = prev }
 }
 
 // parseTime parses a time and returns hours and minutes
--- a/date/date.go
+++ b/date/date.go
@@ -50,7 +50,7 @@ func parseTime(s string) (hour, minute int, err error) {
 	return hour, minute, nil
 }
 
-var TimeFormats = []string{"20060102", "01/02/06"}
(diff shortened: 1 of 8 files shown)


date/date.go
```diff
@@ -43,47 +53,77 @@ func parseTime(s string) (hour, minute int, err error) {
 	return hour, minute, nil
 }
 
-var TimeFormats = []string{"20060102", "01/02/06"}
+var TimeFormats = []string{"20060102", "01/02/06", "01/02/2006"}
 
-// DateParamToEpoch turns a passed string parameter into a unix epoch
-func DateParamToEpoch(s, qtz string, d int64, defaultTimeZone *time.Location) int64 {
```
carrieedwards, line 49: Since this is an exported method, it might be worth keeping it and adding a comment saying it is deprecated, in case its being imported externally.

npazosmendez: Good call. I've restored `DateParamToEpoch` and dropped `ParseAtTimeOr` entirely, as it just shuffles the arguments. I don't love the dual naming of `DateParamToEpoch` vs `ParseAtTime`, but I prefer that over deprecating the public function only because of the name

expr/functions/timeSlice/function.go
```diff
@@ -27,18 +29,37 @@ func New(configFile string) []interfaces.FunctionMetadata {
 	return res
 }
 
+// parseTimeArg parses using date.ParseAtTime and falls back
+// to parsing as an interval for retrocompatibility.
+func parseTimeArg(s string) (int64, error) {
+	if epoch, err := date.ParseAtTime(s, "", time.UTC); err == nil {
```
carrieedwards: I think we want to avoid hardcoding time.UTC here, and use the default time zone instead

npazosmendez: Done

carrieedwards (approved):
LGTM!

PR #932, opened 2026-05-28
fix(parser): correctly handle `|`, `=`, whitespace and multi-byte runes in metric names with a wide `RangeTables`

npazosmendez:
## What

`|`, whitespace and the `=` kwarg separator were only reserved when `RangeTables` was empty. Widening it (e.g. `unicodeRangeTables: ["Common"]`) let them be swallowed into metric names, breaking pipes, whitespace-separated tokens and `key=value` args.

This fix reserves them unconditionally in `parseName`

`TestParseExpr` now runs under a widened `RangeTables` (all scripts) and failed for several cases before the fix. I added a trailing `\n` case for whitespace swallowing and a multi-byte non-ASCII case (`métric.ñame`) for the rune-corruption bug below.

Without the fix, 16 cases fail. Each operator gets absorbed into the name:

`=` swallowed

```
--- FAIL: TestParseExpr/func(metric,_key='value')
    unexpected character
    string_to_parse=`(metric, key='value')`, character_number=14, character=`'`
```

`|` swallowed

```
--- FAIL: TestParseExpr/...requestsHandled|keepLastValue()
    got  Target() = "company.server*.applicationInstance.requestsHandled|keepLastValue"
    want Target() = "keepLastValue"
```

whitespace swallowed

```
--- FAIL: TestParseExpr/foo.bar\n.baz\t
    got  Target() = "foo.bar\n.baz\t"
    want Target() = "foo.bar"
```

multi-byte rune corrupted (only the first byte of each rune is kept)

```
--- FAIL: TestParseExpr/métric.ñame
    got  Target() = "m\xc3tric.\xc3ame"
    want Target() = "métric.ñame"
```

## How

- `|` and whitespace are disallowed from metric names entirely, matching graphite web behavior.
- `parseName` now stops entirely when detecting a `key=value` separator, instead of consulting whether the char (`=`) is allowed
- `parseName` only wrote the first byte of each rune while advancing by the full rune width, corrupting multi-byte names; it now emits the continuation bytes.

### DIFF
--- a/expr/expr_test.go
+++ b/expr/expr_test.go
@@ -172,6 +172,20 @@ func TestEvalExpr(t *testing.T) {
 			from:     1437127020,
 			until:    1437127140,
 		},
+		"metric unicode name inside a function": {
+			metric:  "1Метрика",
+			request: "summarize(1Метрика,'1min')",
+			metricRequest: parser.MetricRequest{
+				Metric: "1Метрика",
+				From:   1437127020,
+				Until:  1437127140,
+			},
+			values:   []float64{343, 407, 385},
+			isAbsent: []bool{false, false, false},
+			stepTime: 60,
+			from:     1437127020,
+			until:    1437127140,
+		},
 	}
 
 	parser.RangeTables = append(parser.RangeTables, unicode.Cyrillic)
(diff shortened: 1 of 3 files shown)


carrieedwards (approved):
LGTM

PR #939, opened 2026-07-02
feat(parser): accept hr/hrs, wk/wks, yr/yrs interval unit aliases

npazosmendez:
Fix `IntervalString` to match other abbreviations for durations, like `hr` for hours.

Python Graphite matches these and many more actually (see [`getUnitString`](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/attime.py#L192)), because it only looks at the first letter of the string. That felt too loose to replicate to me, so I am just adding a few other reasonable cases.


ywwg:
> first letter of the string.

jfc

ywwg (approved):

carrieedwards (approved):

PR #944, opened 2026-08-05
fix(render): make graph functions work in non-cairo builds

npazosmendez:
Most Graphite Web's graph functions (`color()`, `threshold()` etc) "work" with non-image formats (e.g. `format=json`) by being _mostly_ no-op (depends on the function). For example, with `color(randomWalk('a'), 'red')`:

```
$ docker run -d --name gw -p 8080:80 graphiteapp/graphite-statsd >/dev/null && until curl -sf -o /dev/null "http://localhost:8080/render?target=x&format=json"; do sleep 2; done && curl -sG http://localhost:8080/render --data-urlencode "target=color(randomWalk('a'), 'red')" -d format=json -d from=-5min; echo; docker rm -f gw >/dev/null

[{"target": "a", "tags": {"name": "a"}, "datapoints": [[0, 1785953431], [0.24124975089820433, 1785953491], [-0.03474284243293879, 1785953551], [-0.47080131891509025, 1785953611], [-0.3820169792063769, 1785953671]]}]
```

Carbonapi follows that behavior, but only when built with `cairo`. When built without cairo, they return no data (or an error in some cases).

This PR changes that, and allows such functions to run in no-cairo builds. Only the image generation bit (which depends on the `format=` requested) remains under the cairo build tag. If built without cairo, an error is returned, but only when an image is requested explicitly.

**Behavior changes:**
- Non-cairo builds: `color`, `stacked`, `areaBetween`, `alpha`, `dashed`, `drawAsInfinite`, `secondYAxis`, `lineWidth`, `threshold`, `verticalLine` and `holtWintersConfidenceArea` now evaluate normally and appear in `/functions`.
- Non-cairo builds: `format=png`/`svg` returns 501 instead of a 200 with an empty body. Notably, since png is the default format, requests without `format` also get 501 now.
- Piggybacking fix: `dashed()` now honors its dash length argument; a shadowed variable made it always 0, so lines never actually dashed.

**Other changes:**
- Refactor: resolves a pending TO-DO of splitting the graph functions into their own structs, following the common structure. 
- CI now runs the existing `test_nocairo`
- **Perf**: because the `GraphOptions` struct does not depend on the build tag, **`MetricData` type size grows in no-cairo builds from 192 bytes to 272 bytes.** I think this is acceptable and will probably be negligible compared to the memory that comes from tags and sample values. Still I wanted to state it explicitly


expr/functions/cairo/png/png.go
```diff
@@ -4,19 +4,12 @@
...
 	"github.com/go-graphite/carbonapi/expr/types"
-	"github.com/go-graphite/carbonapi/pkg/parser"
 )
 
-const HaveGraphSupport = false
-
-func EvalExprGraph(ctx context.Context, eval interfaces.Evaluator, e parser.Expr, from, until int64, values map[parser.MetricRequest][]*types.MetricData) ([]*types.MetricData, error) {
-	return nil, nil
-}
+const HaveRenderSupport = false
```
carrieedwards: We might want to deprecate the HaveGraphSupport const but leave it in case it is being important in external code

npazosmendez: I thought this was harmless because `HaveGraphSupport` was effectively unused. But I guess I can imagine someone importing this and using the variable to check if the build includes cairo. 

I've reverted back the name to `HaveGraphSupport`, but I should also point out that there are a number of other public interfaces that are being deleted or changing, like `png.EvalExprGraph()`. Honestly I don't think they are worth keeping, it is unlikely that someone is relying on those, and if they are it should be easy to adapt to the new interfaces. These are mostly internal-ish interface changes, not breaking behavior changes

carrieedwards: Fair enough; I think that's likely fine.

